PR #1, opened 2015-03-05
dynamic rest: first commit

aleontiev:


dynamic_rest/viewsets.py
```diff
@@ -0,0 +1,31 @@
...
+      if field.startswith('-'):
+        field = field[1:]
+        include = False
+      else:
+        include = True
+      field_segments = field.split('.')
+      num_segments = len(field_segments)
+      current_map = field_map
+      for i, segment in enumerate(field_segments):
+        if segment:
```
ryochiji: I'm guessing this is so if we get `a..b` we pretend it's `a.b`, but I wonder if it might be better to fail (or ignore the field entirely) since it could be indicative of a bug/issue upstream. Also, weird things might happen if we get `a..` since `a` should be a leaf but won't be treated as such.

aleontiev: Good point; I was intending this to handle `a.b.` but I will change it so that it explicitly fails if there is any empty segment that is not the last segment (i.e. `a..b` and `a.b..`)

aleontiev: I also have not yet includes tests for the viewset, so I will make sure to cover that negative case during the tests.

dynamic_rest/renderers.py
```diff
@@ -0,0 +1,93 @@
...
+          seen = False
+          self._seen[model].add(pk)
+
+        # prevent sideloading the primary objects
+        if depth == 0:
+          return
+
+        # if the primary resource is embedded, add it to a prefixed key
+        if model == self._plural_name:
+          model = '+%s' % model
```
aleontiev: This is outside of the spec; not really sure what the right behavior is. Perhaps we should just prevent this from happening... its pretty hard to actually get 2 serializers to use each other as sub-serializers anyway.

ryochiji: This seems like a reasonable solution. An alternative might be to construct the path. I can imagine a scenario where we do something like `/alos/?fields[]=cards.alos.` in which case I might expect to see the 2nd layer keyed by `cards.alos`. 
On a separate but related note, I wonder if we should limit the depth to something sane...

dynamic_rest/renderers.py
```diff
@@ -0,0 +1,93 @@
...
+    # populated by _sideload
+    self._data = {}
+
+    # used by _sideload to prevent duplicates
+    self._seen = defaultdict(set)
+
+    # used by _sideload to prevent adding secondary records of the primary resource
+    # from ending up in the primary slot
+    self._plural_name = getattr(
+        renderer_context.get('view').get_serializer().Meta, 'plural_name', 'objects')
```
ryochiji: We do this in so many places. I wonder if we should just have a `get_plural_name()` method in one of our Serializer base classes (in `DynamicModelSerializer` maybe?). That might also make it easier to support transitioning from `many_resource_name` to `plural_name` (or maybe even make it optional to explicitly set `plural_name` and return `name + 's'` by default).

dynamic_rest/renderers.py
```diff
@@ -0,0 +1,93 @@
...
+
+
+class DynamicJSONRenderer(renderers.JSONRenderer):
+
+  def render(self, data, accepted_media_type=None, renderer_context=None):
+    # if data is a string, just render it out using the default method
+    # this usually indicates an error message
+    if isinstance(data, basestring):
+      return super(DynamicJSONRenderer, self).render(data, accepted_media_type, renderer_context)
+
```
ryochiji: I think there should be another way to bypass this renderer (for example, for the output of metadata()). Also, some errors (from DRF?) are dicts that look like `{"detail":<error_message>}`. Maybe pass through if data is a `dict` but has no `_model`/`_pk` fields?

dynamic_rest/renderers.py
```diff
@@ -0,0 +1,93 @@
+from collections import defaultdict
+from rest_framework import renderers
+
+
+class DynamicJSONRenderer(renderers.JSONRenderer):
+
+  def render(self, data, accepted_media_type=None, renderer_context=None):
```
ryochiji: BTW, I copied this to Apollo to integrate it with the V3 stuff, and it worked pretty seamlessly. Yay! The only thing I had to change was to check for `_model` so it wouldn't try to modify options responses and batch create responses (see comment below).

dynamic_rest/renderers.py
aleontiev: In the case that `data` is a dict or a list, we will waste a bit of time traversing through the structure looking for `_pk`s and `_model`s but given that an error dict is shallow, this shouldn't be much of a problem; in the end, we will call the superclass anyway to render the same result. Essentially, `_sideload` already does this check.

dynamic_rest/renderers.py
aleontiev: That makes sense -- I'll add `_get_plural_name` -- going to stick with the convention of prefixing our extension methods with `_` to maintain a clear distinction between DRF and our add-ons.

dynamic_rest/renderers.py
aleontiev: Yeah this is an interesting case... I don't really think we should try to optimize for this until we see a user need, because I can see this going one of several ways and it will take some magic to get serializers to be able to reference each other (i.e. there is a circular dependency issue that happens when two serializers attempt to use each other as fields). `cards.alos` makes sense to me. However, it will make things somewhat more difficult for the consumer in the case that the primary resource appears as a secondary in multiple places, e.g. for `GET /alos/?fields[]=cards.alos.&field[]=foo.alos.`, should we return `cards.alos` and also `foo.alos`? The relationships are already in the response, so the consumer should be able to derive them anyway, in which case separate keys might just be a parsing complication.

Let's wait for someone on the frontend to request this and see what they think the behavior should be (my guess is that Alex won't want to support this at all :) ) -- I'll leave a TODO that we should spec the behavior thoroughly.

dynamic_rest/renderers.py
aleontiev: Hmm, perhaps we can gate the actually sideloading logic with a flag on the viewset for OPTIONS? The hacky yet simplifying part of this solution is that the serializers have to "tag" their data with `_model` and `_pk` such that the renderer can sideload recursively without knowing anything about the request. We probably always want to have the renderer strip these tags out, whether there is any actual sideloading or not.

An interesting case to consider though.. I've actually used sideloading in POSTs and PUTs for the `/applicants/` API in vishnu. For example, when creating an event, I use `POST /events/?sideload[]=users` (old syntax) -- the reason being that the event that comes back has a `creator_id` (user) which in some cases I have never sent to the backend and therefore isnt in the Ember store! So Ember will break unless I sideload the user data along with the POST confirmation response.

dynamic_rest/renderers.py
aleontiev: Hmm, actually you're right, I'll think this through some more. We are using this class not only to sideload but also to "emberify" the response such that we key the data by the singular/plural name of the resource, which we dont need for certain response types.

aleontiev: This renderer actually does 4 things right now:

1) strips out `_model` and `_pk` tags added by serializers
2) modifies response such that it is keyed by the serializer resource name
3) sideloading
4) adds `meta` into the response (for pagination)

I am thinking that we should always try to do (1) and allow the viewset to control (2) (3) and (4) -- i.e. you should be able to have a response without `meta` (if the viewset has no meta), with embedded rather than sideloaded records (if the viewset disables sideloading, on by default), and without the Ember-ified structure (if the viewset disables it). And really, we can make (1) an option as well for debugging purposes :)

aleontiev: Considering the way you do `errors` in batch create, it might actually be helpful to extend (4) in such a way that the viewset can provide a map of additional data to the renderer/response that it outside of the realm of serializers and should not be modified, e.g. `viewset.extra_data = { 'errors' : ......, 'meta': ...... }` or `return Response(serializer.data, extra_data = {'errors' .... 'meta' .... }`

dynamic_rest/viewsets.py
aleontiev: Per feedback from the team, I am going to split out `fields[]` into two separate parameters: `include[]` and `exclude[]` such that they can be toggled individually (there seems to be a lot of pushback around exclusions) -- this should not change the underlying map being built though, just the parsing logic

aleontiev:
@ryochiji 
- added tests for `_get_request_fields` including the invalid `a..b` case
- added `_get_plural_name` method and more serializer tests
- added `_sideload` option to viewset such that it can disable the sideloading logic in the renderer
- doc commented all new methods and properties

still TODO:
- renderer tests
- integration tests that hit the DRF router -> DYN viewset -> DYN serializers and DYN renderer (mostly to serve as a model for implementations)

dynamic_rest/renderers.py
ryochiji: Yeah, I'm good with deferring this. It did get me wondering how Ember Data loaded sideloaded objects (a casual search returned a lot of conflicting information) and that might be a factor too...

aleontiev:
@ryochiji added integration tests and a dummy set of test models and settings that will allow us to `make server` in order to bring up a demo server. I think it would be nice if we get this up on Heroku, add some fixture data, and allow folks to play around with the requests via the DRF browsable API.

Will also work on a coverage report; at this point I think it is around 90%

aleontiev:
Having second thoughts about the need to underscore all of the new methods, given that there are certain methods which I would like to consider "private" in the sense that implementations of dynamic-rest will not need to know about them, and others that should be overridable. Perhaps we should make the public-facing methods and properties look more public.

Thoughts on that?

ryochiji:
:+1: for having a demo server.

As for underscores, yeah, I think the underscore-as-private-method convention is so deeply engrained that using it to denote something else is probably not worth the confusion it can cause. What were your concerns around distinguishing Dynamic REST methods and DRF methods?

aleontiev:
My concern is mostly internal w.r.t. Dynamic REST; I would like to continue to stay up-to-date with DRF changes and I felt that drawing a clear distinction between the DRF core and the dynamic extensions would help us do that.

However, from an external perspective, consumers of the Dynamic REST interfaces should know what is "public" and what is "private" and I think that should probably take precedence :)

As such, I'll remove the underscores from the "public" methods

aleontiev:
At this point, I think `include[]` and `exclude[]` are well-supported and tested in DREST at this point, so I am going to merge the PR after making the underscore changes unless you have any other issues. 

Here are a few next-steps I can see, in (IMO) priority order:

1) Port over the prefetching features from `V3` and extend them to support `Prefetch` object chaining which should be useful for filtering down related objects and deferring fields at the queryset level.

2) Port over the `filters{}` feature from `V3`, perhaps modifying the implementation slightly such that it is compatible with the `filter_backends` interface (https://github.com/tomchristie/django-rest-framework/blob/master/rest_framework/generics.py#L42)

3) Add support for pagination, perhaps leveraging the new DRF 3.1 pagination interface (https://github.com/tomchristie/django-rest-framework/blob/master/rest_framework/generics.py#L45)

4) Port over `V3RelationField` and support lazy-loading in order to deal with circular dependencies within serializers (e.g. `users = RelationField('UserSerializer')`)

@ryochiji how is your allocation looking for this sprint? Would love to have your help with any of those points or others we can come up with :) -- but if you are tasked with other work, I have another week dedicated to this so I can start to work on them solo. Lets sync up today to discuss.

ryochiji:
@aleontiev - Yep, looks good to me -- go ahead and merge. I scheduled some time later for us to sync up as well.

PR #2, opened 2015-04-01
add DynamicField/DynamicRelationField and switch to Google Python Style Guide comments

aleontiev:


aleontiev:
@ryochiji very light implementation of `V3RelationField` for now -- left out some of the optimizations. 

One  feature you might be interested in is being able to specify a related serializer by name without the module path, using `self.parent.__module__` as a fallback in case the full module path is not specified.

aleontiev:
Moving this to https://github.com/AltSchool/dynamic-rest/pull/3/files

PR #78, opened 2016-02-06
Revert "global serializer cache"

ryochiji:
Reverts AltSchool/dynamic-rest#72

### DIFF
--- a/.isort.cfg
+++ b/.isort.cfg
@@ -1,3 +1,2 @@
 [settings]
 multi_line_output=3
-known_third_party=django,pylru,rest_framework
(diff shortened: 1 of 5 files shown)


ryochiji:
@aleontiev - Can we merge this? I realized global serializer caching just isn't going to work because context is dynamic, but [DRF caches it](https://github.com/tomchristie/django-rest-framework/blob/master/rest_framework/fields.py#L556) at the instance level. So when serializers are re-used across requests, serializers could be using the wrong cached context.

aleontiev:
:+1:, sorry about the delay

PR #79, opened 2016-02-10
Bulk creation mixin

d2rk:
* Bulk resource creation: `[{"name": "John"}, {"name": "Bob"}]` or `{"users": [{"name": "John"}, {"name": "Bob"}]}`
* Improve `Makefile`
* Add default resource name and plural name


d2rk:
Still fighting with

```
benchmarks/test_bench.py, line 235
  def bench(self):
    fixture 'self' not found
```

aleontiev:
@d2rk don't worry about the `benchmarks` failure; benchmarks are not run as part of the test suite and this is a pre-existing error that's related to the way we are auto-generating the benchmarks as test cases.

dynamic_rest/viewsets.py
```diff
@@ -336,5 +336,21 @@ def list_related(self, request, pk=None, field_name=None):
...
+    """
+    def create(self, request, *args, **kwargs):
+        if not isinstance(request.data, list):
+            return super(BulkCreationMixin, self).create(
+                request, *args, **kwargs)
+        else:
+            serializer = self.get_serializer(data=request.data, many=True)
+            serializer.is_valid(raise_exception=True)
+            self.perform_create(serializer)
+            return Response(serializer.data, status=status.HTTP_201_CREATED)
```
aleontiev: What happens if there a partial failure, e.g. one out of 3 posted object failed?

requirements.txt
```diff
@@ -1,6 +1,7 @@
 Sphinx==1.3.4
 dj-database-url==0.3.0
 flake8==2.4.0
+inflection==0.3.1
```
aleontiev: You should add this to `install_requires.txt`, not `requirements.txt`.

`requirements.txt` is for development requirements (running dynamic-rest as an app or running tests/`make` commands)
`install_requires.txt` is for library requirements. When other apps import dynamic-rest as a dependency, only the dependencies specified in `install_requires.txt` will be installed (via `python setup.py install`)

dynamic_rest/viewsets.py
```diff
@@ -336,5 +336,21 @@ def list_related(self, request, pk=None, field_name=None):
         return Response(serializer.data)
 
 
+class BulkCreationMixin(object):
+    """
+    Either create a single or many model instances in bulk by using the
+    Serializer's many=True ability from Django REST >= 2.2.5.
+    """
+    def create(self, request, *args, **kwargs):
+        if not isinstance(request.data, list):
```
aleontiev: In DREST, we typically support creation of a single option in two different formats, e.g the basic form that is supported by REST framework:

```
{
  "name": "Fido",
  "age": 2
}
```

but also the form where data is keyed by the serializer name (mostly to support Ember Data, but also because this is a more flexible format that allows for metadata to be passed in request data without conflicting with field names):

```
{
  "dog": {
    "name": "fido",
    "age": 2
  }
}
```

Can you modify this check so that it respects the 2nd form, which in this case should use the plural name of the serializer:

```
{
  "dogs": [{
    "name": "Spot"
  }, {
    "name": "Fido"
  }]
}
```

aleontiev:
@d2rk in general, this is an awesome feature to add -- I wouldn't even mind mixing this in to our default ViewSet and perhaps using a viewset flag to toggle this behavior (e.g. ENABLE_BULK_CREATION, defaulting to True)

Take a look at the test fails, it seems related to linting (`isort` is complaining that dependencies are not properly sorted)

requirements.txt
d2rk: Got it, thanks! Fixed.

aleontiev: No problem, this is definitely not intuitive. I'll add some comments to requirements.txt making this clear.

dynamic_rest/viewsets.py
d2rk: Done. However, `get_model`, `get_name` and `get_plural_name` methods have to be class methods then.

d2rk:
Done. This can be done now per view or per application:

```python
class DoggyView(DynamicModelView):

  ENABLE_BULK_CREATION = False
```

or

```python
settings.DYNAMIC_REST['ENABLE_BULK_CREATION'] = False
```

dynamic_rest/viewsets.py
```diff
@@ -337,4 +337,19 @@ def list_related(self, request, pk=None, field_name=None):
 
 
 class DynamicModelViewSet(WithDynamicViewSetMixin, viewsets.ModelViewSet):
-    pass
+
+    ENABLE_BULK_CREATION = dynamic_settings.get('ENABLE_BULK_CREATION', True)
```
aleontiev: Use the special settings wrapper instead: `from dynamic_rest.conf import settings`

d2rk: Hm, I did this in the very same way as the rest of things done: https://github.com/AltSchool/dynamic-rest/blob/master/dynamic_rest/viewsets.py#L70-L71

aleontiev: Ah, don't worry about it then; the settings wrapper was recently introduced to make it easier to document our settings. I must have missed this file during the conversion. I can go over it later and change the syntax

d2rk:
@aleontiev: this is very weird. Still cannot fix this problem. `make tox`, `make tests` and `make lint` are all green. From circleci and can see the only problem `ERROR: py27-lint: commands failed` but not sure how to fix it.

dynamic_rest/serializers.py
```diff
@@ -83,6 +84,16 @@ def __new__(cls, *args, **kwargs):
             meta = type('Meta', (), {})
             cls.Meta = meta
         meta.list_serializer_class = DynamicListSerializer
+
+        # Check for bulk resource creation with help of resource plural name.
+        # This should happen before actual constructor because class needs to
+        # be changed on DynamicListSerializer.
+        plural_name = cls.get_plural_name()
+        data = kwargs.get('data', fields.empty)
+        if data is not fields.empty and plural_name in data and len(data) == 1:
```
aleontiev: This is really cool -- if we're doing this here, do we even need the BulkCreateMixin? If we expand this check so that it looks for list-type values, we can make all DREST serializers support bulk creation:

```
if data is not fields.empty:
  if plural_name in data and len(data) == 1:
    kwargs['data'] = data[plural_name]
    kwargs['many'] = True
  elif isinstance(data, list):
    kwargs['many'] = True
```

trisrael:
This is awesome. :) I'm gonna start looking for ember side of things. Would
this work for UPDATE as well?

aleontiev:
@trisrael don't think it would work for update (PUT/PATCH) as is, but its probably not too far off from it.

@d2rk still interested in the partial failure case; what sort of response should a client expect if 1 out of 3 POSTed records could not be created?

aleontiev:
@d2rk regarding the Circle errors, it looks like an `isort` error:

```
Running isort code checking
ERROR: /home/ubuntu/dynamic-rest/dynamic_rest/serializers.py Imports are incorrectly sorted.
ERROR: /home/ubuntu/dynamic-rest/tests/test_viewsets.py Imports are incorrectly sorted.
isort failed: Some modules have incorrectly ordered imports. Fix by running `isort --recursive .`
```

Currently, `make lint` is somewhat out of date; we should update it to execute `runtests.py --lintonly`, which will run the flake8 check but also a new isort test. This is what is happening in the `py27-lint` test (see `tox.ini`).

I just ran `isort --recursive dynamic_rest tests` and it automatically fixed the imports. In this case, it wasn't actually an ordering issue, it was whitespace between sections. Isort expects imports to fall into 3 sections that are split by a newline: system dependencies first (importlib, json, copy), then 3rd party dependencies (django, rest_framework, inflection), then 1st party dependencies (dynamic_rest).

Here's what I changed: https://github.com/AltSchool/dynamic-rest/commit/87ddc04abab0168485aff8578240c6cd92150f81

dynamic_rest/serializers.py
d2rk: Heh, not that simple, because you're getting loop because of `many_init`: https://github.com/tomchristie/django-rest-framework/blob/master/rest_framework/serializers.py#L104-L134
But I will consider this approach. It would be really nice to have.

d2rk:
Ah, I see. Didn't use this before. Thanks!

dynamic_rest/serializers.py
d2rk: OK, so I removed `ENABLE_BULK_CREATION` from settings and make it default behaviour. I also moved the logic from `__new__` to `DynamicModelViewSet.create`.

ryochiji:
@trisrael - I think one reason we haven't used batch creation from Ember is because we didn't have a good answer for matching up id-less objects being created with the created objects that get returned. Would be nice to get that sorted out though...

d2rk:
@aleontiev, answering your question regarding partial failure case. Unfortunately, there is no native way to disable validators. There will be quite a bit of work to write workaround. Right now it will return 400 in case one of the items already exists. Test was added.

aleontiev:
400 sounds fine, but is there a way for the client to identify which items have failed from the response?

d2rk:
Yes, but already has some details in response, e.g. `'name', ['Group with this name already exists.']`. In case you would like to change this we will have to rework validation process. But it depends what you would like to have in the response. Can you provide an example?

aleontiev:
POSTing a single record will return either a success or error response. On success, the response will include the representation of the created record, including the record ID. On failure, the response will indicate error information as you indicated.

We should try to maintain parity in the multi-POST scenario, i.e. make sure that a client can identify, for each of the POSTed objects, whether the object was created, what the object's ID was if created, and what the error was if not created.

In other words, something like this:

```
>

POST /users/
{
  "users":  [{
    "name": "Joe Smith",
    "age": 20
  }, {
    "name": "Jim Raynor",
    "age": 25
  }, {
    "name": "Joe Smith",
    "age": 26
  }]
}

< 
4XX (bad request?)
{
  "users": [{
    "id": 1,
    "name": "Joe Smith",
    "age": 20,
  }, {
    "id": 2,
    "name": "Jim Raynor",
    "age": 25
  }, {
    "id": null,
    "errors": {
      "name": ["This name is already taken!"]
    }
  }]
}
```

ryochiji:
An alternative format might be something like:

```
{
  "users" : [
    {
      "id": 1,
      ...
    },
    ...
  ],
  "errors": [
    {
      "messages": ["Invalid foo", ...],
      "code": 400,
      "request": {...}
    },
    ...
  ]
}
```

Returning the request object for each would be helpful for the client if there are multiple errors.

I've also been thinking of eventually building a batching proxy that would have an interface similar to FB's batch API: https://developers.facebook.com/docs/graph-api/making-multiple-requests

aleontiev:
@ryochiji with that format, how would you know which object succeeded and which failed?

:+1: to the facebook approach. For clarity, we should try to make a bulk request/response for N objects look as much like N single request/responses.

ryochiji:
@ant - with my proposed format, each failed object will have a separate entry under the `errors` key (so `request` will be one request object, not the whole request payload).

aleontiev:
What if there are two objects in the request that have identical payloads, the first of which succeeds, and the second of which fails because of some uniqueness constraint?

ryochiji:
Yeah, maintaining ordering is helpful for that case. I'm not sure how I feel about mixing errors in with the good objects though -- I feel like that would require clients to scan all returned data in a common deserializer in order to detect errors.

d2rk:
@aleontiev, @ryochiji : I think following http://jsonapi.org/format/#errors should solve this. I've already implemented it and it looks as follows:

```python
{
  'groups': [
    {
      'group': {
          'users': [],
          'name': u'bar',
          'loc1usersLambda': [], 'members': [], 'loc1users': [], 'id': 2, 'permissions': []
      }
    }
  ]
  'errors': [
    {
      'detail': "{'name': [u'Group with this name already exists.']}",
      'source': {u'name': u'foo'}
    }
  ],
}
```

Thoughts?

ryochiji:
That looks similar to what I proposed above (I'm in favor of using JSON API error objects though). I think @aleontiev's concern was that if there are two identical objects and one fails, there's no way to disambiguate. Having said that, even if we order responses, that won't help the client if it isn't aware of ordering (and with Ember Data, I believe the client doesn't know the ordering).

aleontiev:
Cool; lets go with this solution. In general, it would be nice if there was a clear correspondence between the request objects and response objects/errors, but it sounds like thats not a practical concern right now. 

@d2rk can you modify your test for the bulk create partial error case to assert that the response looks as you've indicated in your comment (so that it is clear to developers what kind of output they can expect). After that, let's merge this

vaebnkehn:
@ryochiji - could be solved by the client passing a temporary id and getting back from the POST request a mapping of temporary ids to real ids

d2rk:
@aleontiev, done. Added an extra option `ENABLE_BULK_PARTIAL_CREATION` (default: disabled) that allows you to enable/disable partial creation which means whether or not entries that were successfully verified be created even if there are some errors.

dynamic_rest/viewsets.py
```diff
@@ -337,4 +337,91 @@ def list_related(self, request, pk=None, field_name=None):
 
 
 class DynamicModelViewSet(WithDynamicViewSetMixin, viewsets.ModelViewSet):
-    pass
+
+    ENABLE_BULK_PARTIAL_CREATION = (
+        dynamic_settings.get('ENABLE_BULK_PARTIAL_CREATION', False))
```
aleontiev: The setting that you added to `dynamic_rest.conf` won't be reflected here unless you use the new form for settings:

```
from dynamic_rest.conf import settings
...
ENABLE_BULK_PARTIAL_CREATION = settings.ENABLE_BULK_PARTIAL_CREATION
```

There also seems to be a mismatch of defaults; you have default of False here, but default of True in `conf.py`.

aleontiev:
@d2rk the new flag and implementation looks great! Just one minor comment about settings mismatch around the new flag that's a bit confusing.

dynamic_rest/viewsets.py
d2rk: `False`/`True` typo has been fixed. But I'm not sure about `settings` problem since we already discussed this. I just followed the rest of the code, e.g. `PAGE = dynamic_settings.get('PAGE_QUERY_PARAM', 'page')`. Do you want me to replace `from django.conf import settings` on `from dynamic_rest.conf import settings`?

aleontiev: Adjusted this here: https://github.com/AltSchool/dynamic-rest/commit/da093dbc85d2d458d8f7e28f4d8ed4f25dcbf389

d2rk:
@aleontiev: LGTM.

dynamic_rest/viewsets.py
```diff
@@ -337,4 +336,90 @@ def list_related(self, request, pk=None, field_name=None):
...
+            else:
+                if self.ENABLE_BULK_PARTIAL_CREATION:
+                    self.perform_create(serializer)
+                    items.append(serializer.data)
+                else:
+                    serializers.append(serializer)
+        if not self.ENABLE_BULK_PARTIAL_CREATION and not errors:
+            for serializer in serializers:
+                self.perform_create(serializer)
+                items.append(serializer.data)
```
aleontiev: I noticed there's a bit of redundancy in the output that's caused by the use of `serializer.data`, which wraps the result of `serializer.to_representation`:

```
{
  'groups': [
    {
      'group': {
          'users': [],
          'name': u'bar',
          'loc1usersLambda': [], 'members': [], 'loc1users': [], 'id': 2, 'permissions': []
      }
    }
  ]
  'errors': [
    {
      'detail': "{'name': [u'Group with this name already exists.']}",
      'source': {u'name': u'foo'}
    }
  ],
}
```

Rather than returning a payload that includes "groups" at the top-level, and also the "group" singular name for each group, we should just return "groups" with an array, as we would if we were doing a GET request for groups. 

In other words, it should look like this:

```
{
  'groups': [
    {
        'users': [],
        'name': u'bar',
        'loc1usersLambda': [], 'members': [], 'loc1users': [], 'id': 2, 'permissions': []
    }
  ]
  'errors': [
    {
      'detail': "{'name': [u'Group with this name already exists.']}",
      'source': {u'name': u'foo'}
    }
  ],
}
```

You can accomplish this by replacing `serializer.data` with `serializer.to_representation(serializer.instance)`

d2rk: Done.

aleontiev:
Nice work!

PR #80, opened 2016-02-10
always use distinct() on prefetch queries

ryochiji:
The current approach of deciding when to use `.distinct()` can result in prefetch queries returning multiple rows of the same object (current known example is on a many-to-many field through a join table). This change forces all prefetch queries to use `.distinct()` (which is probably a bit ham-fisted, but should also be safe and low-impact).

See #ALTOS-5703

### DIFF
--- a/dynamic_rest/filters.py
+++ b/dynamic_rest/filters.py
@@ -454,11 +454,13 @@ class DynamicFilterBackend(BaseFilterBackend):
             requirements: An optional TreeMap of nested requirements.
         """
 
+        is_root_level = False
         if serializer:
             if queryset is None:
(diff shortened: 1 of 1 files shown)


aleontiev:
:+1:, nice catch. Will have to look more closely at prefetch querysets; I'm surprised that they weren't getting picked up by the `has_joins` predicate, as a prefetch of B related to A should look like:

```
SELECT *** FROM B INNER JOIN A ON B.ID = A.B_ID WHERE A.ID IN (1, 2, 3, ....)
```

vaebnkehn:
nice one!

PR #84, opened 2016-02-19
support for serializer fields in `requires` for consistent prefetching behavior

ryochiji:
This change adds support for specifying `requires` fields as serializer fields, rather than model fields. In cases where serializer fields are used, the prefetcher treats it identically to field inclusion (e.g. it'll use `serializer.filter_queryset()`). This change maintains backward compatibility so it's still possible to use model fields in `requires`, ~~but in the long run, we should probably deprecate that and only support serializer fields~~.

For context: https://github.com/AltSchool/vishnu-backend/pull/2396

TODO:
- [x] add tests


aleontiev:
Not sure I agree that we should try to deprecate model field requires in the long run; there are some internal fields that might not be exposed in the API but that are required for a computed field that is exposed.

ryochiji:
Yeah, good point. It feels weird to support both, but it seems like there are valid usecases for both.

aleontiev:
I see DREST as a mapping between two inherently separate layers: the REST API and the Django ORM. To the outside world, the REST API consists of resources and resource fields -- the underlying source of these resources is of little importance. But to an API developer, it is valuable to understand both the resources and the models, particularly when the mapping between them is thicker than a rewrite.

I don't think we should try to hide the internal layer from API developers; instead, I would rather that we clearly define the boundary. I agree that supporting both model fields and API fields in the `requires` argument is confusing, and there is no precedence within the existing code for mixing the two concepts as far as I understand.

But maybe I'm missing something; can you elaborate on the underlying need that prompted this change?

ryochiji:
The vishnu-backend PR linked in the description might give some context, but basically, we want the prefetch query built through `requires` to go through `serializer.filter_queryset()` the same way it happens when that relation is sideloaded/included.

dynamic_rest/filters.py
```diff
@@ -433,6 +438,30 @@ def _extract_requirements(
...
+                continue
+            if k in all_fields:
+                # Add this field to list of fields we want to include.
+                # Note that this doesn't change whether or not the field
+                # is included in output because that's controlled by
+                # `serializer.fields`, which we're not modifying.
+                fields[k] = all_fields[k]
+            else:
+                pass
+                #TODO: We could enforce a check here so that `requires`
```
vaebnkehn: lint failure

dynamic_rest/filters.py
```diff
@@ -433,6 +438,30 @@ def _extract_requirements(
...
+                #      keys have to be known serializer fields.
+
+
```
vaebnkehn: lint failure

aleontiev:
Got it.. yeah, that's a tricky problem. The only alternative I can see to this solution is introducing a new dedicated parameter for requirements expressed in API language, but I don't think there would ever be a situation where you would want both API-requirements and model-requirements on the same computed field (how would we even resolve that?).

As such, :+1: to moving ahead with this solution.

ryochiji:
@aleontiev - Added tests and docs.

vaebnkehn:
@ryochiji - what do you need to get this in? This is blocking me

PR #85, opened 2016-02-27
new router API to allow DREST to map/generate canonical resource URLs

ryochiji:
This PR adds a new `router.register_resource()` API for registering the canonical resource URL. It automatically generates the path using the viewset's serializer's resource name to enforce our existing naming convention. The router then internally maintains a reverse map so we can take a resource name and generate a URL with confidence, as well as determine a resource's canonical serializer.

TODO:

- [x] Tests
- [ ] Update docs

### DIFF
--- a/dynamic_rest/routers.py
+++ b/dynamic_rest/routers.py
@@ -143,8 +143,12 @@ class DynamicRouter(DefaultRouter):
 
         # Try to extract resource name from viewset.
         try:
-            resource_name = viewset.serializer_class().get_plural_name()
+            serializer = viewset.serializer_class()
+            resource_key = serializer.get_resource_key()
+            resource_name = serializer.get_plural_name()
         except:
+            import traceback
+            traceback.print_exc()
             raise Exception(
                 "Failed to extract resource name from viewset: '%s'."
                 " It, or its serializer, may not be DREST-compatible." % (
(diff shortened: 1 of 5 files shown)


dynamic_rest/routers.py
```diff
@@ -128,6 +129,87 @@ def register(self, prefix, viewset, base_name=None):
...
+                " It, or its serializer, may not be DREST-compatible." % (
+                    viewset
+                )
+            )
+
+        # Construct canonical path and register it.
+        if namespace:
+            namespace = namespace.rstrip('/') + '/'
+        base_path = namespace or ''
+        base_path = r'%s' % base_path + resource_name
```
aleontiev: The mix of `+` and `%` is a little confusing, how about:

```
base_path = r'%s%s' % (base_path, resource_name)
```

dynamic_rest/routers.py
```diff
@@ -128,6 +129,87 @@ def register(self, prefix, viewset, base_name=None):
...
+        self.register(base_path, viewset)
+
+        # Make sure resource isn't already registered.
+        if resource_name in resource_map:
+            raise Exception(
```
aleontiev: How do you envision this working with multiple versions/namespaces?

aleontiev:
:+1: this approach looks good. I'm curious as to what this might look like as new API versions/namespaces are introduced; is the idea to only register the "latest" version using the new method?

dynamic_rest/routers.py
ryochiji: For now, I think the idea would be that only one can be mapped as canonical using `register_resource`, though it's possible to register other endpoints using the plain `register` API (i.e. both `/v2/foo/` and `/v3/foo/` can exist, but only one can be canonical). It may become necessary to support canonical resources within versions/namespaces, but in our case, we had `/v2` and `/v3` co-exist for a while, so I think this approach would've worked.

ryochiji:
@aleontiev - Updated. Main changes since the last time you looked are:

 * Resources are keyed by `model._meta.db_table` for model-based resources, rather than `serializer.get_name()`. This was because, to support polymorphic relations, I need to be able to look up the serializer given an arbitrary model instance.
 * Added tests for new DynamicRouter methods to get canonical paths and serializers.

aleontiev:
:+1: LGTM. Feel free to merge when ready

PR #87, opened 2016-03-01
Fix/recursive prefetch

ryochiji:
The change in `requires` behavior in https://github.com/AltSchool/dynamic-rest/pull/84 surfaced a Django bug (also see https://github.com/AltSchool/vishnu-backend/pull/2396#issuecomment-190592145) that crops up when the prefetch tree is recursive.

This PR further modifies `requires` to support a workaround in scenarios where this bug is likely to be triggered. This is a really hacky solution, but I'd rather not dive back into the Django prefetch code right now...

### DIFF
--- a/tests/serializers.py
+++ b/tests/serializers.py
@@ -144,6 +144,7 @@ class UserSerializer(DynamicModelSerializer):
             'display_name',
             'thumbnail_url',
             'number_of_cats',
+            'number_of_katz',
             'number_of_alive_cats',
             'profile'
         )
@@ -170,6 +171,10 @@ class UserSerializer(DynamicModelSerializer):
         requires=['location.cat_set.*'],
         deferred=True
     )
+    number_of_katz = DynamicMethodField(
+        requires=['location__cat_set'],
+        deferred=True
+    )
     number_of_alive_cats = DynamicMethodField(
         requires=['location.cats.*'],
         deferred=True
@@ -181,6 +186,9 @@ class UserSerializer(DynamicModelSerializer):
 
     def get_number_of_cats(self, user):
         return len(user.location.cat_set.all())
+
+    def get_number_of_katz(self, user):
+        return self.get_number_of_cats(user)
 
     def get_number_of_alive_cats(self, user):
         # NOTE: This field requires the `location.cats` serializer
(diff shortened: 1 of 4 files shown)


tests/test_prefetch.py
```diff
@@ -33,3 +41,31 @@ def test_nested_prefetch(self):
                 )
             )
         )[0]
+
+    def test_recursive_prefetch_query_bug(self):
```
ryochiji: This test demonstrates the bug in Django

tests/test_prefetch.py
```diff
@@ -33,3 +34,31 @@ def test_nested_prefetch(self):
...
+                )
+            )
+        ))
+
+        users = list(locations[0].user_set.all())
+
+        # NOTE: this is a bug. All the data should be prefetched, but this
+        #       will cause 2 queries to be executed
+        with self.assertNumQueries(2):
+            for user in users:
```
aleontiev: Can you add a comment with the actual queries being executed?

tests/serializers.py
```diff
@@ -170,6 +171,10 @@ class Meta:
         requires=['location.cat_set.*'],
         deferred=True
     )
+    number_of_katz = DynamicMethodField(
+        requires=['location__cat_set'],
```
aleontiev: Hmm, so this seems like it should be equivalent to `requires=['location.cat_set']`. Is that right? Does this issue persist if we use the form: `requires=['location.cat_set.*']`?

ryochiji: Yeah, so what happened with the recent change is that the `location` and `cat_set` are now separated, so we build a Prefetch on Location which then prefetches `cat_set`. Before, it was basically `User.object.prefetch_related('location__catset')` so there was no separate explicit prefetch on Location. [This test](https://github.com/AltSchool/dynamic-rest/pull/87/files#diff-625efbbf35a2290861f596fee1b19838R934) shows how these two fields have different behaviors.

aleontiev: Got it. I thought it might be related to [this recent Django ticket](https://code.djangoproject.com/ticket/25546) but wasn't able to get the test to fail after patching in the fix to that issue.

This is pretty unfortunate.. even with this change, we'd see issues with requests like:

```
GET /users?filter{name}=Foo&include[]=locations.users.cats.
```

ryochiji:
Reverting https://github.com/AltSchool/dynamic-rest/pull/84 so closing this as well.

PR #88, opened 2016-03-02
Revert "support for serializer fields in `requires` for consistent prefetching behavior"

ryochiji:
Reverts AltSchool/dynamic-rest#84


ryochiji:
Reverting the `requires` change since it's breaking too many things in vishnu. If we really need it, I'll try re-implementing it in a way that doesn't break `requires`.

PR #89, opened 2016-03-10
Fix ALTOS-4880

d2rk:
Test that invalid data filtering (by date in this case) returns 400 instead of 500.


ryochiji:
I'm confused... I don't see any changes to DREST, though the test seems to pass. What was the fix?

d2rk:
@ryochiji: sorry, it's not done yet.

d2rk:
@ryochiji: ready now.

Makefile
```diff
@@ -115,11 +118,12 @@ start: install
 	$(INSTALL_DIR)/bin/python manage.py runserver $(PORT) --settings=tests.settings
 
 # Lint the project
-lint: clean_working_directory
+lint: clean_working_directory $(INSTALL_DIR)/bin/flake8
 	$(call header,"Linting code")
-	@find . -type f -name '*.py' -not -path '$(INSTALL_DIR)/*' -not -path './docs/*' -not -path './build/*' | xargs flake8
+	@find . -type f -name '*.py' -not -path '$(INSTALL_DIR)/*' -not -path './docs/*' -not -path './build/*' | xargs $(INSTALL_DIR)/bin/flake8
+	@. $(INSTALL_DIR)/bin/activate; isort --check-only --recursive $(APP_NAME) tests
```
aleontiev: Its better to be consistent about developer dependencies here; we shouldn't need to explicitly have a step for installing flake8. Instead, we can just call the activate script and then use flake8, same way you are doing for isort here

dynamic_rest/filters.py
```diff
@@ -517,7 +518,15 @@ def _filter_queryset(
...
         if query:
-            queryset = queryset.filter(query)
+            # Convert internal django ValidationError to APIException-based one
+            # in order to resolve validation errors from 500 status code to
+            # 400.
+            try:
+                queryset = queryset.filter(query)
+            except InternalValidationError as e:
+                raise ValidationError(
+                    dict(e) if hasattr(self, 'error_dict') else list(e)
```
aleontiev: Should this be `hasattr(e, 'error_dict')`?

tests/test_api.py
```diff
@@ -556,12 +558,21 @@ def test_get_wiht_filter_nested_rewrites(self):
...
         self.assertEqual(400, response.status_code)
 
+    def test_get_with_filter_invalid_data(self):
+        User.objects.create(
+            name='test',
+            date_of_birth=datetime.datetime.utcnow()
+        )
+        url = '/users/?filter{date_of_birth.gt}=0&filter{date_of_birth.lt}=0'
+        response = self.client.get(url)
+        self.assertEqual(400, response.status_code)
```
aleontiev: Can you add an assertion for the response? Would be nice to know what sort of error message you would see if this happened

dynamic_rest/filters.py
d2rk: Yes, see https://github.com/django/django/blob/master/django/core/exceptions.py#L167-L170 for details. This is report `__str__` method because we need actual messages.

aleontiev: Right; but here it says `hasattr(self, 'error_dict')`, where `self` is the `DynamicFilterBackend` object.

Makefile
d2rk: Well, I moved it to virtualenv because right now I always have an error due to not installed flake8 and I don't want to install it globally (e.g. `make lint` doesn't work unless use have flake8). Second, I've been always having a problem with linting because `isort` is not part of it. Not sure what is the problem having these two in virtualenv.

aleontiev: No issues having them in virtualenv, I'm just saying that we already have them in virtualenv without doing any extra steps, because both flake8 and isort are in our requirements.txt file.

As long as the `$(INSTALL_DIR)` step has been run, both flake8 and isort will be accessible, either by calling out to `$(INSTAL_DIR)/bin/flake8` or by calling `$(INSTALL_DIR)/bin/activate` and then using `flake8` without a path perfix (activate will set up your path to point first to the virtualenv)

tests/test_api.py
d2rk: Done.

Makefile
d2rk: The confusion comes from line 79, where you do `lint` before `install` so that `flake8` is not available yet. Thus I added manually. Will fix it then and remove manual flake8 installation.

dynamic_rest/filters.py
d2rk: Good catch. Fixed.

Makefile
aleontiev: Ahh, makes sense. Yeah, it would be better to `install` first to avoid this inconsistency.

aleontiev:
@d2rk nice work, looks great!

PR #91, opened 2016-03-17
support for generic relation fields

ryochiji:
Builds on router changes in https://github.com/AltSchool/dynamic-rest/pull/85 to add improved support for ContentType fields.

**Supported:**
 * Including/sideloading of polymorphic relations using canonical serializer (if registered)
 * Prefetching (some weirdness around how Django seems to fetch/cache ContentType objects themselves)

**Not supported:**
 * `include[]` and `exclude[]` of fields within related objects (which arguably doesn't make sense anyway)
 * `filter{}` for fields within related objects
 * `filter{}` for ContentType fields themselves, other than with `.isnull` operator

**Output structure (proposed):**
(`user.pet` is the polymorphic relation)

```
{
  "users": [{
    "id": 1,
    "pet": {
      "type": "cats",
      "id": 1
    }
  },{
    "id": 2,
    "pet": {
      "type": "dogs",
      "id": 1
    }
  }],
  "cats": [{
    "id": 1,
     ...
  }],
  "dogs": [{
    "id": 1,
     ...
  }]
}
```

**Todo:**
 - [ ] Ratify approach and whether or not this makes sense to add
 - [ ] Ratify output format
 - [ ] When sideloading, parent object should include `type` instead of just the ID
 - [ ] Add some sensible default behavior for when there's no registered serializer for sideloaded resource.
 - [ ] Support for polymorphic many relations?

### DIFF
--- a/dynamic_rest/content_type_field.py
+++ b/dynamic_rest/content_type_field.py
@@ -1,0 +1,56 @@
+import traceback
+
+from dynamic_rest.routers import DynamicRouter
+from dynamic_rest.fields import DynamicField
+
+
+class DynamicContentTypeField(DynamicField):
+
+    def __init__(
+        self,
+        object_field=None,
+        type_field=None,
+        id_field=None,
+        *args, **kwargs
+    ):
+        # NOTE: object_field should be same as `source`, except `source`
+        #       isn't reliably available until this field is bound().
+        if not object_field:
+            raise Exception(
+                'DynamicContentTypeField requires `object_field` kwarg'
+            )
+
+        # Infer type field and id field from object_field if not explicitly set
+        type_field = type_field or object_field + '_type'
+        id_field = id_field or object_field + '_id'
+
+        # Inject `requires` so required fields get prefetched properly.
+        # TODO: It seems like we should be able to require the type and
+        #       id fields, but that seems to conflict with some internal
+        #       Django magic. Disabling `.only()` by requiring '*' seem
+        #       to work more reliably...
+        kwargs['requires'] = kwargs.pop('requires', []) + [
+            object_field + '.*',
+            '*'
+        ]
+        kwargs['read_only'] = True
+        super(DynamicContentTypeField, self).__init__(*args, **kwargs)
+
+    def to_representation(self, instance):
+        try:
+            # Fetch related object, determine its type
+            resource_key = instance._meta.db_table
+
+            # Find serializer for that resource type
+            serializer_class = DynamicRouter.get_canonical_serializer(
+                resource_key
+            )
+
+            # TODO: In theory, it is possible to inject require_fields here
+            #       and support field inclusion/exclusion...
+            return serializer_class(context=self.context).to_representation(
+                instance
+            )
+        except:
+            traceback.print_exc()
+            return None
--- a/dynamic_rest/content_type_field.py
+++ b/dynamic_rest/content_type_field.py
@@ -36,15 +36,32 @@ class DynamicContentTypeField(DynamicField):
         kwargs['read_only'] = True
         super(DynamicContentTypeField, self).__init__(*args, **kwargs)
 
+    def id_only(self):
+        if self.parent and self.field_name:
+            parent_request_fields = getattr(
+                self.parent,
+                'request_fields',
+                {}
+            )
+            return parent_request_fields.get(self.field_name)
+
     def to_representation(self, instance):
+
         try:
-            # Fetch related object, determine its type
+            # Determine resource type
             resource_key = instance._meta.db_table
 
             # Find serializer for that resource type
             serializer_class = DynamicRouter.get_canonical_serializer(
                 resource_key
             )
+
+            # TODO: Add support for `id_only()` here.
+            if self.id_only():
+                return {
+                    'type': serializer_class().get_plural_name(),
+                    'id': instance.pk
+                }
 
             # TODO: In theory, it is possible to inject require_fields here
             #       and support field inclusion/exclusion...
--- a/dynamic_rest/content_type_field.py
+++ b/dynamic_rest/content_type_field.py
@@ -1,7 +1,7 @@
 import traceback
 
-from dynamic_rest.routers import DynamicRouter
 from dynamic_rest.fields import DynamicField
+from dynamic_rest.routers import DynamicRouter
 
 
 class DynamicContentTypeField(DynamicField):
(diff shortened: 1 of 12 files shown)


ryochiji:
CC @alexgb

install_requires.txt
```diff
@@ -1,3 +1,4 @@
 Django>=1.7,<=1.9
 djangorestframework>=3.1.0,<=3.3.0
 inflection==0.3.1
+django-debug-toolbar==1.4
```
aleontiev: Developer requirements should go into `requirements.txt`, which is only referenced when running the development server/tests. `install_requires.txt` requirements will be pulled into all projects that require DREST.

dynamic_rest/content_type_field.py
```diff
@@ -0,0 +1,73 @@
+import traceback
+
+from dynamic_rest.fields import DynamicField
+from dynamic_rest.routers import DynamicRouter
+
+
+class DynamicContentTypeField(DynamicField):
```
aleontiev: Nit on the name of this class: our current `DynamicRelationField` is basically a wrapper around a Django `ForeignKey`/`ManyToMany`/`RelatedObject` relationship; why not create a `DynamicGenericRelationField` that wraps `GenericForeignKey`?

dynamic_rest/content_type_field.py
```diff
@@ -0,0 +1,73 @@
...
+        *args, **kwargs
+    ):
+        # NOTE: object_field should be same as `source`, except `source`
+        #       isn't reliably available until this field is bound().
+        if not object_field:
+            raise Exception(
+                'DynamicContentTypeField requires `object_field` kwarg'
+            )
+
+        # Infer type field and id field from object_field if not explicitly set
```
aleontiev: When would you want to provide an override for `type_field` and `id_field`, rather than inferring from the GenericForeignKey field?

dynamic_rest/content_type_field.py
```diff
@@ -0,0 +1,73 @@
...
+        type_field = type_field or object_field + '_type'
+        id_field = id_field or object_field + '_id'
+
+        # Inject `requires` so required fields get prefetched properly.
+        # TODO: It seems like we should be able to require the type and
+        #       id fields, but that seems to conflict with some internal
+        #       Django magic. Disabling `.only()` by requiring '*' seem
+        #       to work more reliably...
```
aleontiev: Interesting, nice call-out. I'm pleasantly surprised that `prefetch_related` is even compatible with `GenericForeignKey` :)

dynamic_rest/content_type_field.py
```diff
@@ -0,0 +1,73 @@
+import traceback
```
aleontiev: Packaging: since we already have serializer fields avaialble at `dynamic_rest.fields`, we may want to turn `fields` into a directory and make this class live at `dynamic_rest.fields.content_type:DynamicContentTypeField`

dynamic_rest/content_type_field.py
```diff
@@ -0,0 +1,73 @@
...
+            # TODO: Add support for `id_only()` here.
+            if self.id_only():
+                return {
+                    'type': serializer_class().get_plural_name(),
+                    'id': instance.pk
+                }
+
+            # TODO: In theory, it is possible to inject require_fields here
+            #       and support field inclusion/exclusion...
+            return serializer_class(context=self.context).to_representation(
```
aleontiev: Does this relationship field support embedding? What would that look like?

dynamic_rest/content_type_field.py
```diff
@@ -0,0 +1,73 @@
...
+        #       Django magic. Disabling `.only()` by requiring '*' seem
+        #       to work more reliably...
+        kwargs['requires'] = kwargs.pop('requires', []) + [
+            object_field + '.*',
+            '*'
+        ]
+        kwargs['read_only'] = True
+        super(DynamicContentTypeField, self).__init__(*args, **kwargs)
+
+    def id_only(self):
```
aleontiev: This seems different from our `DynamicRelationField` approach, wherein we check `self.request_fields`, which is populated earlier. Can we consolidate, or is there a reason that we can't pass `request_fields` down to this field at binding time?

tests/test_content_type_field.py
```diff
@@ -0,0 +1,84 @@
...
+        )
+        self.assertTrue('cats' in content)
+        self.assertEqual(2, len(content['cats']))
+        self.assertTrue('dogs' in content)
+        self.assertEqual(1, len(content['dogs']))
+
+    def test_query_counts(self):
+        # NOTE: Django doesn't seem to prefetch ContentType objects
+        #       themselves, and rather caches internally. That means
+        #       this call could do 5 SQL queries if the Cat and Dog
```
aleontiev: Interesting; could this cause N+1 query characteristics, or would each distinct ContentType object only be fetched once?

aleontiev:
Exciting stuff :+1:

tests/test_content_type_field.py
ryochiji: Each distinct ContentType object seems to get fetched once and then cached. I tried sideloading the ContentType objects themselves, but Django still fetches them one by one (and also does a prefetch query, so it's worse)...

dynamic_rest/content_type_field.py
ryochiji: I was able to add embedding pretty easily, but I think the structure needs to be a little different to denote the object type. It could look something like:

```
{
  "user": {
    ...
    "pet": {
      "type": "cats",
      "data": {
        "id": 1,
        ...
    }
  }
}
```

or maybe...

```
{
  "user": {
    ...
    "pet": {
      "cat": {
        "id": 1,
        ...
    }
  }
}

File: dynamic_rest/content_type_field.py (hunk 3 above)
Comment (ryochiji) [inline dynamic_rest/content_type_field.py]: Yeah, this is mostly a side-effect of trying to do this in `__init__` where `parent` isn't available yet, so the model field isn't accessible. I think doing all this in `bind()` is pretty safe, since fields should be bound before any of the other methods in this class get called. I'll try moving it there and see how that looks.

File: dynamic_rest/content_type_field.py (hunk 6 above)
Comment (aleontiev) [inline dynamic_rest/content_type_field.py]: Could also in-line the type field as a "meta" field:

```
{
  "user": {
    ...
    "pet": {
        "id": 1,
        "name": "miao",
        "_type": "cat"
        ...
    }
  }
}
```

ryochiji: @trisrael - Some discussion here of what the JSON structure should be for polymorphic relations. Lacking any other standard we could adopt, I think we should go with whatever Ember Data currently expects.

PR #92, opened 2016-03-28
Fix/serializer name optimization

ryochiji:
Re-computing the plural name is apparently surprisingly costly. This PR caches the value so we only compute it once.

screen shot 2016-03-28 at 9 50 54 am


aleontiev:
:+1:

PR #94, opened 2016-04-02
Feature/dynamic generic relation field

ryochiji:
Cleaned up version of https://github.com/AltSchool/dynamic-rest/pull/91.

Adds a `DynamicGenericRelationField` to support ContentType-based generic relation fields. When sideloading and including, the "primary key" is replaced with an object that looks like:

```
{
  "type": <type name>,
  "id": <primary key value>
}
```


dynamic_rest/processors.py
```diff
@@ -78,14 +78,18 @@ def process(self, obj, parent=None, parent_key=None, depth=0):
                     return
 
                 name = obj.serializer.get_plural_name()
-                pk = obj.instance.pk
+                pk = obj.pk_value or obj.instance.pk
```
ryochiji: This changes the sideloading behavior to make sure the `pk` is replaced with an object that contains `type` and `id`. Basically DynamicGenericRelationField attaches the desired `pk` value to the `TaggedDict` object (`obj`), which is used here.

tests/serializers.py
```diff
@@ -172,6 +177,7 @@ class Meta:
         'ProfileSerializer',
         deferred=True
     )
+    favorite_pet = DynamicGenericRelationField()
```
ryochiji: Example usage.

dynamic_rest/routers.py
```diff
@@ -260,17 +262,11 @@ class UserSerializer(..):
...
-        fields = getattr(serializer, 'get_all_fields', serializer.get_fields)()
+        fields = getattr(serializer, 'get_link_fields', lambda: [])()
 
         route_name = '{basename}-{methodnamehyphen}'
 
         for field_name, field in six.iteritems(fields):
-            # Only apply to DynamicRelationFields
-            # TODO(ryo): Maybe also apply in cases where fields are mapped
-            #            directly to serializers?
-            if not isinstance(field, DynamicRelationField):
```
ryochiji: Not strictly related to this PR, but in refactoring `fields`, importing DynamicRelationField was causing circular dependencies, and realized using `serializer.get_link_fields()` solved that problem and also made things more consistent.

dynamic_rest/fields/content_type.py
```diff
@@ -0,0 +1,91 @@
+import traceback
```
aleontiev: Maybe call this `dynamic_rest.fields.generic` for consistency with the field name?

dynamic_rest/fields/content_type.py
```diff
@@ -0,0 +1,91 @@
...
+        super(DynamicGenericRelationField, self).bind(field_name, parent)
+
+        source = self.source or field_name
+
+        # Inject `requires` so required fields get prefetched properly.
+        # TODO: It seems like we should be able to require the type and
+        #       id fields, but that seems to conflict with some internal
+        #       Django magic. Disabling `.only()` by requiring '*' seem
+        #       to work more reliably...
+        self.requires = [
```
aleontiev: What if `requires` is specified manually as well? Should we concatenate, or is it safe to over-write here?

dynamic_rest/fields/content_type.py
```diff
@@ -0,0 +1,91 @@
...
+            # rather than just the ID.
+            pk_value = self.get_pk_object(
+                serializer_class().get_plural_name(),
+                instance.pk
+            )
+            if self.id_only():
+                return pk_value
+
+            # Serialize the object. Note that request_fields is set, which
+            # means field inclusion/exclusion *may* work. Yolo!
```
aleontiev: Can you elaborate on this comment?

dynamic_rest/fields/content_type.py
```diff
@@ -0,0 +1,91 @@
...
+            ).to_representation(
+                instance
+            )
+
+            # Pass pk object that contains type and ID to TaggedDict object
+            # so that Processor can use it when the field gets sideloaded.
+            if isinstance(r, TaggedDict):
+                r.pk_value = pk_value
+            return r
+        except:
```
aleontiev: What sort of exceptions could we actually see here?

dynamic_rest/processors.py
```diff
@@ -78,14 +78,18 @@ def process(self, obj, parent=None, parent_key=None, depth=0):
...
+
+                # For polymorphic relations, `pk` can be a dict, so use the
+                # string representation (dict isn't hashable).
+                pk_key = repr(pk)
```
aleontiev: Is a dict always going to `repr`d the same way?

dynamic_rest/fields/content_type.py
```diff
@@ -0,0 +1,91 @@
...
+            serializer_class = DynamicRouter.get_canonical_serializer(
+                resource_key=None,
+                instance=instance
+            )
+            if not serializer_class:
+                # Can't find canonical serializer! For now, just return
+                # object name and ID, and hope the client knows what to do
+                # with it.
+                return self.get_pk_object(
+                    instance._meta.object_name,
```
aleontiev: Why do we use `_meta.object_name` here [_meta.db_table](https://github.com/AltSchool/dynamic-rest/pull/94/files#diff-f072b5566ec48c2be4de15dc12a59eacR221) elsewhere?

dynamic_rest/fields/content_type.py
ryochiji: Yeah, given that the behavior for `required` on these fields is pretty fixed, it's safer to just override. I'll add a check to raise an error if `requires` is set manually so it's not silently ignoring it.

dynamic_rest/processors.py
ryochiji: Given equal values, seems like it, though this is far from exhaustive.

```python
>>> d = {}
>>> d['id'] = 'id'
>>> d['type'] = 'type'
>>> e = {}
>>> e['type'] = 'type'
>>> e['id'] = 'id'
>>> f = {'id':'id', 'type':'type'}
>>> repr(d)
"{'type': 'type', 'id': 'id'}"
>>> repr(e)
"{'type': 'type', 'id': 'id'}"
>>> repr(f)
"{'type': 'type', 'id': 'id'}"
```

One alternative would be to construct a tuple using the `id` and `type` but that would make assumptions about the structure/content of `pk`... Also, I think the only consequence of this hashing failing is that the same object might be sideloaded twice, which I don't think is too bad of a failure mode.

dynamic_rest/fields/content_type.py
ryochiji: This is for the scenario where the canonical serializer couldn't be found. We return an object, and infer the `type` based on the model name, so `type: Zebra` is probably more likely to be useful to the client than `type: app_zebra`.

dynamic_rest/fields/content_type.py
ryochiji: Actually, I don't think inclusion/exclusion would work. Going to remove it and add a check back in to explicitly disallow inclusion/exclusion on generic fields.

dynamic_rest/fields/content_type.py
ryochiji: Not expecting anything, but just a catch-all to not break the request if anything unexpected happens. I'll add a comment, and a TODO to remove the try-block once we have a little more confidence.

dynamic_rest/fields/content_type.py
aleontiev: Do you mean inclusion/exclusion of fields on the generic related object, or the relationship itself? 

If its the former, sounds fine. It might be nice to support inclusion/exclusion of common fields like `id` or `created`, but that's not a deal-breaker.

ryochiji: Yes, the former. Incidentally, if I remove that check, inclusion/exclusion does work... as long as the included/excluded fields exist on the related object (if not, there's an exception thrown in `DGRF.to_representation()`). So I think there's probably a way to support it, but I'm going to punt on it for now.

aleontiev: :+1: thanks for the context

ryochiji:
@aleontiev - Addressed comments if you want to take a last look.

aleontiev:
:+1: lgtm

PR #95, opened 2016-04-08
Feature: DynamicGenericRelationField write flows

DeviateFish:
Follows up on #94, adding the ability to put/patch against a polymorphic field.

I'll be following up with some additional testing, as I'm only currently testing the `PATCH` case.

Depends on #94 

Note that this also changes the assumption around the `_type` properties... Now, they're the singular version of the name, not the plural.

### DIFF
--- a/dynamic_rest/fields/generic.py
+++ b/dynamic_rest/fields/generic.py
@@ -22,7 +22,6 @@ class DynamicGenericRelationField(
 
         super(DynamicGenericRelationField, self).__init__(*args, **kwargs)
         self.embed = embed
-        self.read_only = False
 
     def bind(self, field_name, parent):
         super(DynamicGenericRelationField, self).bind(field_name, parent)
@@ -84,7 +83,7 @@ class DynamicGenericRelationField(
             # We want the pk to be represented as an object with type,
             # rather than just the ID.
             pk_value = self.get_pk_object(
-                serializer_class().get_name(),
+                serializer_class.get_name(),
                 instance.pk
             )
             if self.id_only():
--- a/dynamic_rest/fields/generic.py
+++ b/dynamic_rest/fields/generic.py
@@ -84,7 +84,7 @@ class DynamicGenericRelationField(
             # We want the pk to be represented as an object with type,
             # rather than just the ID.
             pk_value = self.get_pk_object(
-                serializer_class().get_plural_name(),
+                serializer_class().get_name(),
                 instance.pk
             )
             if self.id_only():
@@ -117,13 +117,12 @@ class DynamicGenericRelationField(
         model_name = data.get('type', None)
         model_id = data.get('id', None)
         if model_name and model_id:
-            # model_name = inflection.singularize(model_name)
             serializer_class = DynamicRouter.get_canonical_serializer(
                 resource_key=None,
                 resource_name=model_name
             )
             if serializer_class:
                 model = serializer_class.get_model()
-                return model.objects.get(id=model_id)
+                return model.objects.get(id=model_id) if model else None
 
         return None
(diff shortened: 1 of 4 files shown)


ryochiji:
@DeviateFish - #94 merged. Can you make a new PR against master?

DeviateFish:
@ryochiji  I rebased this on master to capture the changes from #94

dynamic_rest/fields/generic.py
```diff
@@ -22,7 +22,7 @@ def __init__(self, embed=False, *args, **kwargs):
 
         super(DynamicGenericRelationField, self).__init__(*args, **kwargs)
         self.embed = embed
-        self.read_only = True
+        self.read_only = False
```
ryochiji: I think we should just remove this. There may be cases where we'd want to set it on a per-field basis.

dynamic_rest/fields/generic.py
```diff
@@ -84,7 +84,7 @@ def to_representation(self, instance):
             # We want the pk to be represented as an object with type,
             # rather than just the ID.
             pk_value = self.get_pk_object(
-                serializer_class().get_plural_name(),
+                serializer_class().get_name(),
```
ryochiji: Let's change this to `serializer_class.get_name()` since it's a [class method](https://github.com/AltSchool/dynamic-rest/blob/master/dynamic_rest/serializers.py#L222) and it'll save us the overhead of constructing a serializer.

dynamic_rest/routers.py
```diff
@@ -179,6 +181,7 @@ def register_resource(self, viewset, namespace=None):
             'path': base_path,
             'viewset': viewset
         }
+        resource_name_map[resource_name] = resource_key
```
ryochiji: Can we add a check to throw an error if `resource_name` is already there? Probably better to err on the side of not allowing ambiguity.

ryochiji:
Sorry about the slow response on this. Few minor comments, but overall looks great.

dynamic_rest/fields/generic.py
DeviateFish: `get_plural_name` is also a class method, it seems.  I should change that one as well, correct?

ryochiji: Sure, though are you using `get_plural_name` anywhere?

DeviateFish: Oh, looks like the [place I'm using](https://github.com/AltSchool/dynamic-rest/pull/95/files#diff-f072b5566ec48c2be4de15dc12a59eacR150) it I'm already invoking it as a class method

dynamic_rest/routers.py
DeviateFish: Done

dynamic_rest/fields/generic.py
DeviateFish: Done

dynamic_rest/routers.py
```diff
@@ -180,6 +182,24 @@ def register_resource(self, viewset, namespace=None):
             'viewset': viewset
         }
 
+        # Make sure the resource name isn't registered, either
+        # TODO: Think of a better way to clean this up, there's a lot of
```
DeviateFish: I added this `TODO` because I do think we could come up with a better way of doing this.  We're enforcing that both `resource_name` and `resource_key` be unique, and they should have a 1:1 mapping between the two, so I think we could come up with a way to reduce this to a single identifier.

dynamic_rest/fields/generic.py
DeviateFish: Done

dynamic_rest/routers.py
ryochiji: resource_key is the DB table name which contains the app label, so technically it supports two models with the same name, as long as they're in different apps/have different table names. In practice, it's probably 1:1, or close enough that we probably won't lose much from saying it should be.

ryochiji:
👍 - Good to merge once tests pass

PR #96, opened 2016-04-13
Make the prefix for additional primary resources configurable

DeviateFish:
Makes the configurable. Using `_` instead of `+` matches what Ember expects for "additional sideloaded primary resources", and should allow it to use its magic. However, for backwards compatibility, the default will remain `+`

### DIFF
--- a/dynamic_rest/conf.py
+++ b/dynamic_rest/conf.py
@@ -35,7 +35,11 @@ DYNAMIC_REST = {
 
     # PAGE_SIZE_QUERY_PARAM: global setting for the page size query parameter.
     # Can be overriden at the viewset level.
-    'PAGE_SIZE_QUERY_PARAM': 'per_page'
+    'PAGE_SIZE_QUERY_PARAM': 'per_page',
+
+    # ADDITIONAL_PRIMARY_RESOURCE_PREFIX: String to prefix additional
+    # instances of the primary resource when sideloading.
+    'ADDITIONAL_PRIMARY_RESOURCE_PREFIX': '_'
 }
 
 
--- a/dynamic_rest/conf.py
+++ b/dynamic_rest/conf.py
@@ -39,7 +39,7 @@ DYNAMIC_REST = {
 
     # ADDITIONAL_PRIMARY_RESOURCE_PREFIX: String to prefix additional
     # instances of the primary resource when sideloading.
-    'ADDITIONAL_PRIMARY_RESOURCE_PREFIX': '_'
+    'ADDITIONAL_PRIMARY_RESOURCE_PREFIX': '+'
 }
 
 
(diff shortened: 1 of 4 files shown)


ryochiji:
@aleontiev - Any idea why `isort` might complain even though the imports haven't changed?

dynamic_rest/processors.py
```diff
@@ -8,6 +8,9 @@
 from dynamic_rest.tagged import TaggedDict
 
 
+ADDITIONAL_PRIMARY_RESOURCE_PREFIX = '_'
```
ryochiji: Might make sense to make this a setting.

ryochiji:
@DeviateFish - Can you add a quick test as well?

aleontiev:
Is `isort` still an issue? Build looks green at this point

dynamic_rest/conf.py
```diff
@@ -35,7 +35,11 @@
 
     # PAGE_SIZE_QUERY_PARAM: global setting for the page size query parameter.
     # Can be overriden at the viewset level.
-    'PAGE_SIZE_QUERY_PARAM': 'per_page'
+    'PAGE_SIZE_QUERY_PARAM': 'per_page',
+
+    # ADDITIONAL_PRIMARY_RESOURCE_PREFIX: String to prefix additional
+    # instances of the primary resource when sideloading.
+    'ADDITIONAL_PRIMARY_RESOURCE_PREFIX': '_'
```
aleontiev: Can you also add this into the README? We have a section there for configuration

DeviateFish: Done

DeviateFish:
`isort` stopped being an issue when I moved the prefix to a configuration setting, for some reason.

ryochiji:
Yeah, looks like having the module-level constant threw it off. @aleontiev - this was the build that failed: https://circleci.com/gh/AltSchool/dynamic-rest/352

aleontiev:
Ah -- yeah, isort makes some assumptions about spacing in between the import sections and the code section.

DeviateFish:
Note: We'll probably want to set this configuration variable to `+` when we use this change in the backend, until we've finished migrating frontends to use new serializer apis.  We'll want to at least provide parity, to start, until all the frontends are up to speed.

ryochiji:
LGTM.

@aleontiev - we should make of note of this change when release on PyPi since it's a potentially breaking change.

DeviateFish:
Would it make more sense to make the default `+` for parity, then override it in our own applications later?

tests/test_api.py
```diff
@@ -1344,3 +1344,36 @@ def test_get_list(self):
...
+            backup_home_id=backup_home_id
+        )
+        self.kitten = Cat.objects.create(
+            name='Kitten',
+            home_id=home_id,
+            backup_home_id=backup_home_id,
+            parent=parent
+        )
+
+    def test_additional_sideloads(self):
```
ryochiji: Can we add another test where we override that setting, and confirm that it gets used?

ryochiji:
I think leaving it as `+` by default makes sense.

PR #97, opened 2016-04-19
fix for making DynamicRelationField read-only by including in Meta.re…

ryochiji:
Fixes a bug where adding a DynamicRelationField to `Meta.read_only_fields` doesn't actually make it read-only. This happens because DRF doesn't automagically trickle that down for us (it only does it for direct serializer mapping) and we weren't pulling it down in DREST either.

### DIFF
--- a/dynamic_rest/fields.py
+++ b/dynamic_rest/fields.py
@@ -122,6 +122,7 @@ class DynamicRelationField(DynamicField):
             # related_name
             model_field = None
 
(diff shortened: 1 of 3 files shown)


ryochiji:
CC @vaebnkehn

vaebnkehn:
Nice test.  Thanks!

aleontiev:
:+1:

PR #98, opened 2016-04-26
Bulk update support

d2rk:
Implements bulk update support.

Examples:

```
PATCH /dogs/1/
{
    'fur': 'white'
}
```

```
PATCH /dogs/
[
    {'id': 1, 'fur': 'white'},
    {'id': 2, 'fur': 'black'},
    {'id': 3, 'fur': 'yellow'}
]
```

```
PATCH /dogs/?filter{fur.contains}=brown
[
    {'fur': 'grey'}
]
```


dynamic_rest/serializers.py
```diff
@@ -76,6 +80,49 @@ def data(self):
...
+        updated_objects = []
+        for object_to_update in objects_to_update:
+            lookup_key = getattr(object_to_update, lookup_attr)
+            data = lookup_objects.get(lookup_key)
+            # Use model serializer to actually update the model
+            # in case that method is overwritten.
+            updated_objects.append(self.child.update(object_to_update, data))
+
+        if all_objects_update:
+            queryset.update(**all_objects_update)
```
aleontiev: One important caveat to doing a bulk update in this way is that `post_save` signals will not get fired -- its worth either calling that out explicitly in documentation, or calling `.save` in a loop instead.

dynamic_rest/serializers.py
```diff
@@ -76,6 +80,49 @@ def data(self):
                 self._sideloaded_data = ReturnList(data, serializer=self)
         return self._sideloaded_data
 
+    def update(self, queryset, validated_data):
```
aleontiev: Can we get this to respect DREST's request style where the data is keyed by the serializer name? e.g these should both work:

```
PATCH /dogs/
{
    "dogs": [{
        "id": 1,
        "fur": "brown"
    }, {
       "id": 2,
       "fur": "tan"
    }]
}
```
and
```
PATCH /dogs/
   [{
        "id": 1,
        "fur": "brown"
    }, {
       "id": 2,
       "fur": "tan"
    }]
```

aleontiev:
The update-by-query style is interesting, but the interface seems inconsistent:

```
PATCH /dogs/?filter{fur.contains}=brown
[
    {'fur': 'grey'}
]
```

What's with the array notation here? The expectation is that all `dogs` matching the filter will be patched with the same data, so why not use the object notation as in the basic case:

Update a specific `dog` with a specific object:
```
PATCH /dogs/1/
{
    'fur': 'white'
}
```

Update many `dogs` with a specific object:

```
PATCH /dogs/?filter{fur.contains}=brown
 {
    'fur': 'grey'
}
```

Update ALL `dogs` with a specific object:

```
PATCH /dogs/
 {
    'fur': 'grey'
}
```

vs. update some specific `dogs` with specific objects:

```
PATCH /dogs/
[
    {'id': 1, 'fur': 'white'},
    {'id': 2, 'fur': 'black'},
    {'id': 3, 'fur': 'yellow'}
]
```

aleontiev:
Now that I've seen the potential power of the "PATCH all" style of this feature, I do think it would be helpful to leave this functionality off by default and enable it on a view-by-view basis.

I want to make sure we don't accidentally run a bulk operation against an entire resource, when we actually intended to run it against a single instance of that resource.

```
PATCH /groups/      # oops, we forgot to add the group ID to the end of this call!
{
    "name": "foobar"
}
```

d2rk:
Yeah, so it's kind of overkill. I guess we should get rid of "PATCH all" scenario for safety. Even on view-by-view basis user can send such request and it will be very hard for us to limit it. So let's force the user to pull all object IDs first and then PATCH them in one request.

dynamic_rest/serializers.py
d2rk: Done.

+ tests

dynamic_rest/serializers.py
d2rk: Not a case any more.

d2rk:
'PATCH all' has been removed. At some point, you can think of it as 'DELETE all'.

All comments were addressed. Tests were added.

dynamic_rest/viewsets.py
```diff
@@ -338,6 +338,71 @@ def list_related(self, request, pk=None, field_name=None):
...
     ENABLE_BULK_PARTIAL_CREATION = settings.ENABLE_BULK_PARTIAL_CREATION
+    ENABLE_BULK_UPDATE = settings.ENABLE_BULK_UPDATE
+
+    def get_object(self):
+        lookup_url_kwarg = self.lookup_url_kwarg or self.lookup_field
+
+        if lookup_url_kwarg in self.kwargs:
+            return super(DynamicModelViewSet, self).get_object()
+
+        # If the lookup_url_kwarg is not present
```
aleontiev: What is the purpose of this override? Is this to allow `PATCH` to be called without an ID?

aleontiev:
:+1: to removing the PATCH all behavior for now to be safe.

I've got one minor question about the purpose of the `get_object` override; seems like we should be able to avoid that.

d2rk:
Yeap, not necessary anymore. Thanks!

docs/genindex.html
```diff
@@ -281,172 +223,34 @@ <h2 id="D">D</h2>
   </dt>
 
       
-  <dt><a href="dynamic_rest.html#dynamic_rest.fields.DynamicComputedField">DynamicComputedField (class in dynamic_rest.fields)</a>
```
aleontiev: What happened to all of the `dynamic_rest.fields` documentation? Surprised to see it redlined when there are no actual fields changes in this PR

d2rk: For some reason `make docs` was failing with `django.core.exceptions.AppRegistryNotReady`. Fixed this problem and regenerated docs.

aleontiev:
Awesome work! @ryochiji FYI

dynamic_rest/serializers.py
```diff
@@ -407,6 +445,26 @@ def to_representation(self, instance):
             embed=self.embed
         )
 
+    def to_internal_value(self, data):
+        value = super(WithDynamicSerializerMixin, self).to_internal_value(data)
+        id_attr = getattr(self.Meta, 'update_lookup_field', 'id')
+        request_method = getattr(
+            getattr(self.context.get('view'), 'request'),
```
ryochiji: FYI, DRF puts the request in context [by default](https://github.com/tomchristie/django-rest-framework/blob/master/rest_framework/generics.py#L136).

d2rk: Thanks!

PR #100, opened 2016-05-04
Feature/immutable fields

ryochiji:
Adds `immutable` option to DynamicFields. If this flag is set, the field can be set in a POST/create request, but can't be updated in a PUT/PATCH. It works by dynamically setting `read_only` to `True` for non-POST requests.

### DIFF
--- a/dynamic_rest/fields/fields.py
+++ b/dynamic_rest/fields/fields.py
@@ -74,11 +74,7 @@ class DynamicField(fields.Field):
         )
 
         # If immutable and not a POST, set read_only to True
-        request_method = getattr(
-            self.context.get('request'),
-            'method',
-            ''
-        )
+        request_method = self.parent.get_request_method()
         if self.immutable and request_method != 'POST':
             self.read_only = True
 
--- a/dynamic_rest/fields/fields.py
+++ b/dynamic_rest/fields/fields.py
@@ -51,33 +51,6 @@ class DynamicField(fields.Field):
     def to_internal_value(self, data):
         return data
 
-    def get_meta_arg(self, local_kwarg, meta_attr):
-        """ Get arguments that can either be set on the field itself,
-        or on a Meta attribute (e.g. "read_only").
-        """
-
-        # Explicit kwarg on field takes precedence
-        if local_kwarg in self.kwargs:
-            return self.kwargs[local_kwarg]
-
-        # Otherwise infer from parent's Meta attr
-        parent_meta_attr = getattr(self.parent.Meta, meta_attr, [])
-        return self.field_name in parent_meta_attr
-
-    def bind(self, *args, **kwargs):
-        super(DynamicField, self).bind(*args, **kwargs)
-
-        self.read_only = self.get_meta_arg('read_only', 'read_only_fields')
-        self.immutable = self.immutable or self.get_meta_arg(
-            'immutable',
-            'immutable_fields'
-        )
-
-        # If immutable and not a POST, set read_only to True
-        request_method = self.parent.get_request_method()
-        if self.immutable and request_method != 'POST':
-            self.read_only = True
-
 
 class DynamicComputedField(DynamicField):
     pass
--- a/dynamic_rest/fields/fields.py
+++ b/dynamic_rest/fields/fields.py
@@ -68,6 +68,10 @@ class DynamicField(fields.Field):
         super(DynamicField, self).bind(*args, **kwargs)
 
         self.read_only = self.get_meta_arg('read_only', 'read_only_fields')
+        self.immutable = self.immutable or self.get_meta_arg(
+            'immutable',
+            'immutable_fields'
+        )
 
         # If immutable and not a POST, set read_only to True
         request_method = getattr(
--- a/dynamic_rest/fields/fields.py
+++ b/dynamic_rest/fields/fields.py
@@ -26,6 +26,7 @@ class DynamicField(fields.Field):
         requires=None,
         deferred=False,
         field_type=None,
+        immutable=False,
         **kwargs
     ):
         """
@@ -40,6 +41,7 @@ class DynamicField(fields.Field):
         self.requires = requires
         self.deferred = deferred
         self.field_type = field_type
+        self.immutable = immutable
         self.kwargs = kwargs
         super(DynamicField, self).__init__(**kwargs)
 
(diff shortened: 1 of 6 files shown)


tests/test_serializers.py
```diff
@@ -662,6 +663,9 @@ def test_different_roots(self):
             'Different root serializers should yield different instances.'
         )
 
+    @unittest.skip(
```
ryochiji: Calling `self.context` in DynamicField uncovered the fact that DRF's `Field.root` property doesn't do cycle detection so this test was hanging.

dynamic_rest/fields/fields.py
```diff
@@ -49,17 +51,36 @@ def to_representation(self, value):
     def to_internal_value(self, data):
         return data
 
+    def get_meta_arg(self, local_kwarg, meta_attr):
```
aleontiev: Might be reasonable to default `meta_attr` to `local_kwarg + '_fields'` as that seems to be the convention

dynamic_rest/fields/fields.py
```diff
@@ -49,17 +51,36 @@ def to_representation(self, value):
...
-            )
-            self.read_only = self.field_name in parent_ro_fields
+        self.read_only = self.get_meta_arg('read_only', 'read_only_fields')
+        self.immutable = self.immutable or self.get_meta_arg(
+            'immutable',
+            'immutable_fields'
+        )
+
+        # If immutable and not a POST, set read_only to True
+        request_method = getattr(
```
aleontiev: Starting to see this in multiple places, might be helpful to introduce `get_request` at the serializer level

aleontiev:
:+1: looks great, few minor suggestions

dynamic_rest/fields/fields.py
```diff
@@ -49,17 +51,36 @@ def to_representation(self, value):
...
-        # Infer read_only from parent Meta.read_only_fields
-        if 'read_only' not in self.kwargs:
-            parent_ro_fields = getattr(
-                self.parent.Meta,
-                'read_only_fields',
-                []
-            )
-            self.read_only = self.field_name in parent_ro_fields
+        self.read_only = self.get_meta_arg('read_only', 'read_only_fields')
+        self.immutable = self.immutable or self.get_meta_arg(
```
aleontiev: Isn't `self.immutable or self.get_meta_arg('immutable', ...)` redundant? Or is `immutable` not going to be stored in `self.kwargs`?

aleontiev: Nevermind, I see that `self.kwargs` is something that we're doing to capture all DRF-only kwargs

tests/test_api.py
```diff
@@ -1395,3 +1395,36 @@ def test_additional_sideloads(self):
...
+            json.dumps(data),
+            content_type='application/json'
+        )
+        self.assertEqual(201, response.status_code)
+        data = json.loads(response.content.decode('utf-8'))
+        self.assertEqual(data['cat']['parent'], parent_id)
+
+        # Try to change 'parent' in a PATCH request...
+        patch_data = {
+            'parent': self.kitten.pk
```
aleontiev: Should we add `immutable` to our `OPTIONS` response? Currently, the behavior for read-only fields is to ignore any new values that might be passed by the client, which is fine... but at least read-only fields can be identified as such by the client using OPTIONS. With immutable fields, there is currently no way of knowing.

dynamic_rest/fields/fields.py
ryochiji: Done. I didn't originally because it didn't feel like a serializer concern... but probably worth DRYing.

tests/test_api.py
ryochiji: Good idea. Done.

ryochiji:
@aleontiev - Made some pretty [significant changes](https://github.com/AltSchool/dynamic-rest/pull/100/commits/a2efa4ff64ee6f09f7535e8355e5773847f80979) because I realized putting all the logic in DynamicField meant we couldn't support normal DRF fields (and would then be misleading to have an `immutable_fields` meta attr). All the logic is now in the serializer's `get_fields()` method, so it works for both DREST and DRF fields.

dynamic_rest/serializers.py
```diff
@@ -290,6 +290,13 @@ def get_plural_name(cls):
             )
         return cls.Meta.plural_name
 
+    def get_request_method(self):
+        return getattr(
+            self.context.get('request'),
+            'method',
+            ''
+        )
```
aleontiev: maybe call `.upper()` on this just in case the underlying value ever becomes lower-case

dynamic_rest/serializers.py
```diff
@@ -346,6 +361,22 @@ def get_fields(self):
         for name in deferred:
             serializer_fields.pop(name)
 
+        # Set read_only flags based on read_only_fields meta list.
+        # Here to cover DynamicFields not covered by DRF.
+        ro_fields = getattr(self.Meta, 'read_only_fields', [])
+        self.make_fields_read_only(serializer_fields, ro_fields, True)
```
aleontiev: This would be more re-usable if it took a flag name parameter similar to `_get_flagged_field_names`, e.g:

```
self.flag_fields(serializer_fields, ro_fields, 'read_only', True)
```

dynamic_rest/serializers.py
```diff
@@ -346,6 +361,22 @@ def get_fields(self):
...
+
+        # Toggle read_only flags for immutable fields.
+        immutable_field_names = self._get_flagged_field_names(
+            serializer_fields,
+            'immutable'
+        )
+        self.make_fields_read_only(
```
aleontiev: This could potentially set a field that is flagged read-only AND immutable as not read-only.

I'm not sure why you would ever do that in practice, but I can see it happening accidentally from an overdose of serializer mixins.

ryochiji: Actually this was intentional. One case we want this today is with UUID `id` fields, which I believe DRF automatically infers as a read-only field, but in some cases want to allow the client to generate IDs and set on POST. This approach allows us to override the field and make it immutable, without having to explicitly declare the field.

aleontiev: That makes sense. So we're saying that immutable + read-only is an acceptable combination, and immutable will override read-only on POST requests. Sounds reasonable, but it would be nice to have a comment explaining this case.

ryochiji:
@aleontiev - updated

aleontiev:
:+1: looks good!

PR #101, opened 2016-05-04
Add default values for getattr

d2rk:
Fix a few `getattr` related bugs.

### DIFF
--- a/dynamic_rest/serializers.py
+++ b/dynamic_rest/serializers.py
@@ -449,7 +449,7 @@ class WithDynamicSerializerMixin(WithResourceKeyMixin, DynamicSerializerBase):
         value = super(WithDynamicSerializerMixin, self).to_internal_value(data)
         id_attr = getattr(self.Meta, 'update_lookup_field', 'id')
         request_method = getattr(
-            getattr(self.context.get('view'), 'request'),
+            getattr(self.context.get('view'), 'request', None),
             'method',
             ''
         )
@@ -510,7 +510,7 @@ class WithDynamicModelSerializerMixin(WithDynamicSerializerMixin):
 
     @classmethod
     def get_model(cls):
-        return cls.Meta.model
+        return getattr(cls.Meta, 'model', None)
 
     def get_id_fields(self):
         """
--- a/dynamic_rest/serializers.py
+++ b/dynamic_rest/serializers.py
@@ -449,7 +449,7 @@ class WithDynamicSerializerMixin(WithResourceKeyMixin, DynamicSerializerBase):
         value = super(WithDynamicSerializerMixin, self).to_internal_value(data)
         id_attr = getattr(self.Meta, 'update_lookup_field', 'id')
         request_method = getattr(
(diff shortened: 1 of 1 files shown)


dynamic_rest/serializers.py
```diff
@@ -510,7 +510,7 @@ class WithDynamicModelSerializerMixin(WithDynamicSerializerMixin):
 
     @classmethod
     def get_model(cls):
-        return cls.Meta.model
+        return getattr(cls.Meta, 'model', None)
```
aleontiev: What's the manifestation of this bug? If a ModelSerializer has model, things will start to fail down the line..

dynamic_rest/serializers.py
```diff
@@ -449,7 +449,7 @@ def to_internal_value(self, data):
         value = super(WithDynamicSerializerMixin, self).to_internal_value(data)
         id_attr = getattr(self.Meta, 'update_lookup_field', 'id')
         request_method = getattr(
-            getattr(self.context.get('view'), 'request'),
+            getattr(self.context.get('view'), 'request', None),
```
aleontiev: According to Ryo's comment, DRF will add request onto the context by default, so you should just do `getattr(self.context, 'request', None)` instead of reaching into the view

dynamic_rest/serializers.py
d2rk: Yes, if you didn't provide `Meta` class and thus it won't have `model` attribute this will raise `AttributeError`. However, serializer may not have related model.

dynamic_rest/serializers.py
d2rk: Thanks, good to know. Fixed.

dynamic_rest/serializers.py
aleontiev: This would still raise an error if the serializer has no Meta class, since we're calling `cls.Meta`.

d2rk: No, because it will be automatically injected. See https://github.com/AltSchool/dynamic-rest/blob/master/dynamic_rest/serializers.py#L133-L136

dynamic_rest/serializers.py
```diff
@@ -449,7 +449,7 @@ def to_internal_value(self, data):
         value = super(WithDynamicSerializerMixin, self).to_internal_value(data)
         id_attr = getattr(self.Meta, 'update_lookup_field', 'id')
         request_method = getattr(
-            getattr(self.context.get('view'), 'request'),
+            self.context.get('request', None),
```
aleontiev: If this returns `None`, the surrounding `getattr` will fail.

dynamic_rest/serializers.py
aleontiev: Ah, good point

dynamic_rest/serializers.py
d2rk: Sorry, I don't see a problem. `getattr(None, 'method', '')` resolves to `''`.

PR #103, opened 2016-05-09
Feature/auto defer many relations

ryochiji:
Ember Data gets confused when many-relations return IDs, and there is also a link object. This PR adds an option to automatically defer many-relations, unless `deferred=False` is explicitly set on the field. The option is disabled by default for now for backward compatibility.


aleontiev:
:+1:

PR #109, opened 2016-05-13
Allow _extract_filters to be view independent by allowing filters_map

rgangopadhya:
argument to be an empty dict, too (previously checked if filters_map was
falsey, and would check for a view if so).

### DIFF
--- a/dynamic_rest/filters.py
+++ b/dynamic_rest/filters.py
@@ -203,10 +203,9 @@ class DynamicFilterBackend(BaseFilterBackend):
 
         """
 
-        filters_map = (
-            kwargs.get('filters_map') or
-            self.view.get_request_feature(self.view.FILTER)
-        )
+        filters_map = kwargs.get('filters_map')
+        if filters_map is None:
+            filters_map = self.view.get_request_feature(self.view.FILTER)
 
         out = TreeMap()
 


aleontiev:
This looks good to me. @ryochiji any concerns?

dynamic_rest/filters.py
```diff
@@ -203,10 +203,9 @@ def _extract_filters(self, **kwargs):
 
         """
 
-        filters_map = (
-            kwargs.get('filters_map') or
-            self.view.get_request_feature(self.view.FILTER)
-        )
+        filters_map = kwargs.get('filters_map')
+        if filters_map is None:
```
ryochiji: We could be even more defensive and do something like:

```python
filters_map = (
    self.view.get_request_feature(self.view.FILTER) 
    if self.view else {}
)
```

or add

```python
assert self.view or 'filters_map' in kwargs, (
    "`filters_map` required if `view` isn't set"
)
```

dynamic_rest/filters.py
```diff
@@ -203,10 +203,9 @@ def _extract_filters(self, **kwargs):
 
         """
```
ryochiji: @rgangopadhya - Since you wanted to use this as a standalone function, I wonder if we should just move it to a helper function and make this method wrap around the function.

Something like:

```python
def extract_filters(view=None, filters_map=None):
```

PR #111, opened 2016-05-25
link field performance optimization

ryochiji:
The `get_link_fields` method was calling `field.serializer`, causing serializers to be constructed, which turns out to be very expensive. This PR uses alternative means of accomplishing the same logic that don't require serializers to be constructed. (In one API hitting 4 resources, this resulted in a ~25% reduction in CPU usage.)

Before (blue boxes show serializer construction caused by `get_link_fields`):
screen shot 2016-05-25 at 3 09 55 pm

After:
screen shot 2016-05-25 at 4 08 42 pm

### DIFF
--- a/dynamic_rest/serializers.py
+++ b/dynamic_rest/serializers.py
@@ -424,12 +424,12 @@ class WithDynamicSerializerMixin(WithResourceKeyMixin, DynamicSerializerBase):
                     not (
                         # Skip sideloaded fields
                         name in self.fields and
-                        not field.serializer.id_only()
+                        isinstance(self.request_fields.get(name), dict)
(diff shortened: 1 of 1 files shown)


dynamic_rest/serializers.py
```diff
@@ -414,12 +414,12 @@ def get_link_fields(self):
                 not (
                     # Skip sideloaded fields
                     name in self.fields and
-                    not field.serializer.id_only()
+                    isinstance(self.request_fields.get(name), dict)
```
aleontiev: Can you add this logic into a helper like `self.is_field_sideloaded` in case we need to do this again for other meta functionality?

PR #113, opened 2016-05-28
fix bug in GenericFKField

ryochiji:
There was a bug in GenericForeignKeyField's `id_only()` method where it was erroneously returning `False` if the field was being included by default (and therefore didn't have a key in `request_fields`). This PR fixes the bug by using the new `.is_field_sideloaded()` method (i.e. `id_only` is true if the field isn't being sideloaded).


ryochiji:
CC @rgangopadhya

aleontiev:
:+1:

PR #115, opened 2016-06-02
protect against weird and broken query params, e.g. "?" where they sh…

benbardin:
…ouldnt be.

This corrects the response to queries like `http://localhost:9001/v3/posts?filter{creator}=17401?&format=json`, which now returns a 400 with `[
"invalid literal for int() with base 10: '17401?'"
]` instead of a 500.


aleontiev:
:+1: -- can you add a test that proves a query like `?filter{creator}=123x` will now result in a 4XX error?

benbardin:
done

PR #116, opened 2016-06-14
fix bulk-creation sideloading bug

ryochiji:
Fixes bug in bulk-creation where sideloads are returned nested, instead of pulled out under top-level keys. This was happening because the response wasn't being passed through `SideloadingProcessor`.


alexmadjar:
👍  - lgtm   Feel free to merge unless you want @aleontiev's thumb too

aleontiev:
:+1: thanks for the tests

PR #117, opened 2016-06-15
Support Meta.preserve_whitespace fields list

alexmadjar:
DRF automatically trims whitespace on CharFields.  This can be turned off by setting the flag to False on that field.   This PR adds a Meta parameter for overriding this flag.

### DIFF
--- a/dynamic_rest/serializers.py
+++ b/dynamic_rest/serializers.py
@@ -393,6 +393,14 @@ class WithDynamicSerializerMixin(WithResourceKeyMixin, DynamicSerializerBase):
         ro_fields = getattr(self.Meta, 'read_only_fields', [])
         self.flag_fields(serializer_fields, ro_fields, 'read_only', True)
 
+        pw_fields = getattr(self.Meta, 'preserve_whitespace', [])
+        self.flag_fields(
+            serializer_fields,
+            pw_fields,
+            'trim_whitespace',
+            False,
+        )
+
         # Toggle read_only flags for immutable fields.
         # Note: This overrides `read_only` if both are set, to allow
         #       inferred DRF fields to be made immutable.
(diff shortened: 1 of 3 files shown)


dynamic_rest/serializers.py
```diff
@@ -549,7 +557,7 @@ def save(self, *args, **kwargs):
         return instance
 
     def id_only(self):
-        """Check whether the serializer should return an ID instead of an object.
+        """Whether the serializer should return an ID instead of an object.
```
alexmadjar: drive by:  makes the line length < 80

aleontiev:
:+1:, would be good to add some documentation that surfaces this feature, e.g. in a docstring at the serializer layer, which would then be exposed in the generated HTML docs.

alexmadjar:
@aleontiev - I don't see where we document the `Meta` api

aleontiev:
@alexmadjar I recommend adding a section to [the DREST serializer base mixin.](https://github.com/AltSchool/dynamic-rest/blob/master/dynamic_rest/serializers.py#L120)

dynamic_rest/serializers.py
```diff
@@ -393,6 +393,14 @@ def get_fields(self):
         ro_fields = getattr(self.Meta, 'read_only_fields', [])
         self.flag_fields(serializer_fields, ro_fields, 'read_only', True)
 
+        pw_fields = getattr(self.Meta, 'preserve_whitespace', [])
```
ryochiji: I'd prefer to keep the convention of ending these meta fields with `_fields` (just `preserved_whitespace` could look like it should be a boolean flag), but not sure what a non-awkward name would be. Maybe `whitespace_preserved_fields`?

alexmadjar:
nits addressed

ryochiji:
👍

PR #121, opened 2016-08-02
Remove isort testing

aleontiev:
`isort` tests that module imports are sorted in alphabetical order within system, 3rd party, and 1st party blocks.

Unfortunately, it seems to be somewhat flaky in its detection of 3rd party vs 1st party packages, to the extent that it produces inconsistent results between local testing and CircleCI testing, and even between different runs of the same code on CircleCI (??).

As such, we have decided to remove it and rely exclusively on flake8 for linting.

### DIFF
--- a/.isort.cfg
+++ b/.isort.cfg
@@ -1,2 +1,0 @@
-[settings]
-multi_line_output=3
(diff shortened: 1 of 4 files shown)


ryochiji:
👍

PR #123, opened 2016-08-03
version 1.5.0

aleontiev:
- Broaden Django requirements to include 1.10
- Bump Django 1.8/1.9 versions in tox
- Bump DREST to 1.5.0 to prepare for PyPi release


ryochiji:
👍

aleontiev:
@ivlevdenis there were a few issues with Django 1.10 that weren't caught in your initial change because the install requirements at that point did not include 1.10.

Once I broadened the version requirement, I found a few issues that are fixed here, although it should be noted that Django 1.10 is not compatible with DRF versions prior to 3.4

PR #124, opened 2016-08-08
Convert filter values to booleans for boolean fields

ryochiji:
Currently queries like `filter{boolean_field}=false` are broken because we're not converting "false" into the boolean value `False` (we were only doing that conversion on `.isnull` filters). This PR fixes that so we convert values if the field is a `BooleanField` (or `NullBooleanField`).


dynamic_rest/filters.py
```diff
@@ -283,7 +285,9 @@ def _filters_to_query(self, includes, excludes, serializer, q=None):
         def rewrite_filters(filters, serializer):
             out = {}
             for k, node in six.iteritems(filters):
-                filter_key = node.generate_query_key(serializer)
+                filter_key, field = node.generate_query_key(serializer)
+                if isinstance(field, (BooleanField, NullBooleanField)):
+                    node.value = node.value.lower() not in self.FALSEY_STRINGS
```
aleontiev: Should we try to validate/prevent non-boolean-looking values here?

E.g. if I make q

aleontiev: Should we try to validate/prevent non-boolean-looking values here?

E.g. if I make a query like "GET /cats?filter{is_dead}=no", that would be interpreted as True.. but seems like it should throw a 400 instead.

ryochiji: I feel like that might be good in the long term, but has a higher likelihood of breaking existing code?

aleontiev: Thats true. What if we introduce it behind a flag that we keep off for backwards compatibility?

ryochiji: Sure. Would you be ok if I did that in a follow-up PR? There's a P1 bug that's waiting for this fix so I'd like to get this in first.

aleontiev: No problem

aleontiev:
:+1:

PR #127, opened 2016-09-13
Fixed python 3 compatibility

shanx:


shanx:
Will isolate this commit into a separate branch, since I'll be creating multiple pull requests

PR #130, opened 2016-09-14
Made overload of api.html template explicit

shanx:
Made overload of api.html template explicit to fix issues of dynamic rest directory rendering in other projects


aleontiev:
:+1:

PR #131, opened 2016-10-07
Add patch to list route

rgangopadhya:
This is to enable bulk PATCH requests, without this PATCH not allowed.


ryochiji:
Can you update the bulk update test so it's actually hitting the route instead of going straight to the view?

rgangopadhya:
@ryochiji updated

PR #132, opened 2016-10-11
Use script_prefix so SCRIPT_NAME can be used to deploy dynamic rest on a subpath

shanx:


shanx:
Sorry was a bit too enthusiastic with this feature and not performing tests through tox. This is not backward compatible with older django versions. So I'll fix and do a new pull request

PR #133, opened 2016-10-12
Add bulk delete ALTOS-11223

rgangopadhya:
https://altschool.atlassian.net/browse/ALTOS-11223

Not sure if the expected format I documented is desired. Thoughts @ryochiji?

CC - @magistrula


dynamic_rest/viewsets.py
```diff
@@ -468,10 +485,50 @@ def create(self, request, *args, **kwargs):
...
+        ):
+            self.check_object_permissions(self.request, instance)
+            self.perform_destroy(instance)
+        return Response(status=status.HTTP_204_NO_CONTENT)
+
+    def destroy(self, request, *args, **kwargs):
+        """
+        Either delete a single or many model instances in bulk
+
+        DELETE /dogs/
```
ryochiji: I'm not sure we should support deletion of single records this way (i.e. should just be `DELETE /dogs/1/`)

dynamic_rest/routers.py
```diff
@@ -57,16 +57,17 @@ def sort_key(r):
     return directory_list
 
 
-def add_patch_to_list_route(routes):
+def modify_list_route(routes):
```
ryochiji: This should probably be gated on `settings.ENABLE_BULK_UPDATE`.

dynamic_rest/viewsets.py
ryochiji: I'm also not sure the code as it is now supports this. It won't be considered a bulk request and I'm pretty sure `super` won't handle this case either.

magistrula: That constraint would cause some messiness on the FE. Bulk delete is needed to support bulk unassigning student cards in Groupwide Assessment, and it would be nice not to have to care how many students cards happen to be assigned. Is it possible to do a bulk patch if we only pass 1 id?

rgangopadhya: @ryochiji yeah known issue that this case is not currently supported (see failing test and PR description). 

@magistrula Ryo is referring specifically to the first payload, which is not a 'bulk' format. The other formats below would support 1 or more ids.

rgangopadhya: The main reason I added that format was for parity with the bulk POST and PATCH cases which support that already. More than happy to remove as that was the one part that was WIP about this PR.

rgangopadhya:
@ryochiji updated to remove the single deletion case.

rgangopadhya:
Yeah, you're right, removed comment.

dynamic_rest/viewsets.py
```diff
@@ -468,10 +485,45 @@ def create(self, request, *args, **kwargs):
...
+        {
+            "dog": {
```
ryochiji: I think this example should also be removed?

It also might be good to actually detect those cases and return a `MethodNotAllowed` error.

rgangopadhya: @ryochiji 
what is the easy way to know the URL a request is made on (for rejecting certain DELETE situations seems easiest if i reject when I know the request is against list URL). alternatively i could reject any request where the request is not bulk and pk not in kwargs

ryochiji: I'm not sure depending on the URL structure is safe, but the default [`destroy()`](https://github.com/tomchristie/django-rest-framework/blob/master/rest_framework/mixins.py#L92) method is pretty light. One option might be to try to do `self.get_object()` and if that raises, assume it's an unsupported request and return an error. By default it would just be an ISE, I think?

rgangopadhya:
Ok, modified based on that. Ready for review again @ryochiji !

dynamic_rest/viewsets.py
```diff
@@ -340,6 +340,23 @@ class DynamicModelViewSet(WithDynamicViewSetMixin, viewsets.ModelViewSet):
...
 
+    def is_bulk_request(self, request):
+        plural_name = self.get_serializer_class().get_plural_name()
+        if isinstance(request.data, list):
+            return True
+        elif plural_name in request.data and len(request.data) == 1:
+            return True
+        return False
+
+    def _get_bulk_payload(self, request):
```
ryochiji: This and `is_bulk_request` look nearly identical. I think it'd be DRYer if this method just returned `None` if it's not a bulk request, and the code in update/delete was basically:

```python
bulk_payload = self._get_bulk_payload(request)
if bulk_payload:
    # do bulk operation
```

rgangopadhya: 👍 updated, rebased to just one commit

rgangopadhya:
Ready for review again @ryochiji

dynamic_rest/viewsets.py
ryochiji: Let's delete this example from the docstring.

ryochiji:
One comment, but otherwise looks good.

dynamic_rest/viewsets.py
rgangopadhya: Arghhh I really thought I did...

PR #134, opened 2016-10-13
Exclude delete requests from having fields deferred

DeviateFish:
When using `.only` on querysets, the objects returned aren't actually instances of the model, but derived instances of the class that signify that certain fields are deferred.  This causes certain signals to not be fired, since we typically wire these up by way of `post_save` or `post_delete` hooks, connected with a particular model class as a sender.

Since django apparently strictly matches the model being acted on to the sender passed in through the `@receiver` decorator or `.connect` (i.e. `@receiver(post_delete, sender=FooModel)`), this causes models with deferred fields to not invoke the proper signal handlers.

See this [stack overflow answer](See http://stackoverflow.com/questions/35006328/post-delete-pre-delete-signals-not-firing-for-specific-sender) for more details.

This fix prevents the use of `.only` on `DELETE` requests in addition to `PUT/PATCH/POST` requests, which in turn prevents deferred subclasses from being used.

### DIFF
--- a/dynamic_rest/filters.py
+++ b/dynamic_rest/filters.py
@@ -504,7 +504,11 @@ class DynamicFilterBackend(BaseFilterBackend):
         # use requirements at this level to limit fields selected
         # only do this for GET requests where we are not requesting the
         # entire fieldset
-        if '*' not in requirements and not self.view.is_update():
+        if (
+            '*' not in requirements and
+            not self.view.is_update() and
+            not self.view.is_delete()
+        ):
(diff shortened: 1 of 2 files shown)


aleontiev:
:+1: makes sense. I believe Django has a real fix for this issue landing in 1.10, but lets go with this until we can upgrade.

aleontiev:
@DeviateFish feel free to bump and tag

sganeriwal:
Nasty; great catch Benton!

PR #135, opened 2016-10-14
Only return 405 on AssertionError (unable to retrieve pk)

rgangopadhya:
We don't want to return 405's in all situations, only when we weren't able to retrieve the primary key. `get_object` could fail even if we have the primary key, for example if `filter_queryset` excludes the object.


rgangopadhya:
Necessary because when user does not have overlord permissions to access, `get_object()` will raise an Http 404 error.

rgangopadhya:
Updated

ryochiji:
Looks good. Can you add a test for the 404 deletion case?

rgangopadhya:
Done

ryochiji:
👍  - good to merge once tests pass.

PR #137, opened 2016-10-24
Use script_prefix so SCRIPT_NAME can be used to deploy dynamic rest on a subpath

shanx:
Allow dynamic rest to be deployed in a script prefix. So instead of deploying Django on `/` on a hostname deploy on `/some/prefix/` and have Dynamic rest use this prefix to generate url's correctly.

This time backward compatible with django < 1.10.x


aleontiev:
@shanx thanks for this patch!

PR #140, opened 2016-11-10
Fix/138

### LINKED ISSUE #138: return structure of DynamicSortingFilter.get_valid_fields does not match djangorestframework's OrderingFilter.get_valid_fields
hi, 

I found a bug if make a search with a field that does not exists in the serializer : ``/?filter{fieldnotinserializer}=val``. 

ValueError: too many values to unpack

it seem that the signature of DynamicSortingFilter.get_valid_fields does not match his parent class's.
the original result from django-rest-framework is a list of 2n tuple, but you return a list of 3n tuple. 

it seem that this broke the OrderingFilter.get_template_context method since the commit aeb57913c9f2a7ad1ffc1db825c3e6f44b4818ee of restframowork which do no longer pass the result of get_valid_fields but iterate over it instead. 

this happen by browsing the django-rest-framework interface

version used : 

- djangorestframework==3.4.0
- dynamic-rest==1.5.0

aleontiev:
Fixes #138 by removing label from `DynamicSortingFilter.get_valid_fields` return

### DIFF
--- a/README.md
+++ b/README.md
@@ -20,10 +20,10 @@ See http://dynamic-rest.readthedocs.org for full documentation.
   - [Linked relationships](#linked-relationships)
   - [Sideloaded relationships](#sideloaded-relationships)
   - [Embedded relationships](#embedded-relationships)
-  - [Field inclusions](#field-inclusions)
-  - [Field exclusions](#field-exclusions)
-  - [Field-based filtering](#field-based-filtering)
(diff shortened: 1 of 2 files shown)


kciredor:
ping @ryochiji Please merge, I can adopt it in my PR #143 (DRF 3.5 compat) to improve it.

PR #141, opened 2016-11-15
Deprecated django 1.7

shanx:
Deprecated the use of django 1.7 as per discussion here: #129. And only now do I find out that auvipy is not a developer of the project :)

PR #142, opened 2016-11-17
Repair model field access in django 1.10

benbardin:


dynamic_rest/fields/fields.py
```diff
@@ -118,8 +122,15 @@ def bind(self, *args, **kwargs):
         parent_model = getattr(self.parent.Meta, 'model', None)
 
         remote = is_field_remote(parent_model, self.source)
+
+        def get_model_field():
```
benbardin: I saw that, but there's a whole lot of logic I don't quite follow in that helper function. (See the "except" business there). Is that all correct here? I'm not sure, so I thought it safer to use this simpler logic that's identical with previous. Thoughts?

dynamic_rest/fields/fields.py
```diff
@@ -118,8 +122,15 @@ def bind(self, *args, **kwargs):
...
+            if DJANGO110:
+                return parent_model._meta.get_field(self.source)
+            else:
+                return parent_model._meta.get_field_by_name(self.source)[0]
```
auvipy: that can be done with older releases. anyone need 1.7 will use older releases and newer release shouldn't continue to support unmaintained django versions.

dynamic_rest/fields/fields.py
aleontiev: Should be fine to use that logic here; it is intended to retrieve remote fields.

dynamic_rest/fields/fields.py
ryochiji: As you probably saw in #141, we need to support 1.7 for a while longer for our own purposes, but we'll likely drop support in the near future.

PR #145, opened 2016-11-23
fix bug where model field lacks args

benbardin:
1. Repairs broken code introduced in #142 .
2. Alters unit tests to be more flexible regarding nullability of inverse fields. There's precedent for this; we believe that the `get_model_field` upgrade actually uncovers a bug that was previously masked.

### DIFF
--- a/dynamic_rest/fields/fields.py
+++ b/dynamic_rest/fields/fields.py
@@ -120,7 +120,7 @@ class DynamicRelationField(WithRelationalFieldMixin, DynamicField):
         remote = is_field_remote(parent_model, self.source)
 
         try:
-            model_field = get_model_field()
+            model_field = get_model_field(parent_model, self.source)
         except:
             # model field may not be available for m2o fields with no
             # related_name
--- a/dynamic_rest/fields/fields.py
+++ b/dynamic_rest/fields/fields.py
@@ -135,7 +135,7 @@ class DynamicRelationField(WithRelationalFieldMixin, DynamicField):
                 )
         ):
             self.required = False
-        if 'allow_null' in self.kwargs and getattr(
+        if 'allow_null' not in self.kwargs and getattr(
                 model_field, 'null', False
         ):
             self.allow_null = True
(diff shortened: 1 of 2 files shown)


dynamic_rest/fields/fields.py
```diff
@@ -135,7 +135,7 @@ def bind(self, *args, **kwargs):
                 )
         ):
             self.required = False
-        if 'allow_null' not in self.kwargs and getattr(
+        if 'allow_null' in self.kwargs and getattr(
```
ryochiji: What's the intention of this change?

benbardin: Cool, I buy that this is wrong. The problem I'm trying to fix: In the course of moving from the locally-defined get_model_field() in my last PR, to the more complex get_model_field() in meta.py, the unit tests broke. Specifically, a bunch of fields that hadn't been nullable, now became nullable (https://circleci.com/gh/AltSchool/dynamic-rest/534, towards the bottom). I _think_ this is because meta.py's get_model_field is more robust, meaning model_field is more often populated, meaning model_field.null is more often available. So, I wasn't sure if (a) the unit tests are, in fact, wrong, or (b) 'allow_null' was set incorrectly here - as a bug - and that bug was eclipsed by the model_fields that were less populated.

So, I put this back, and change the unit tests? Or are they correct, and the problem is elsewhere?

benbardin: Definitely some funky business going on here. If I re-insert `not in`, then I have some tests in dependents of dynamic-rest failing because model fields marked "null" aren't treated as such.

benbardin: Ah, nm on that - `not in` seems correct to dependents. Just wondering about the failing unit test then.

ryochiji:
Can you add to the description the underlying issue you're addressing in this PR?

aleontiev:
:+1: -- changing unit tests are somewhat troubling but make sense given the comment about Django 1.7 vs 1.9 interpretation of reverse relationship fields' nullable status. 

I believe that's actually uncovered a bug in the previous logic where-in we only looked at local fields to determine nullable status, i.e:

```
model_field = parent_model._meta.get_field_by_name(self.source)[0]
```

benbardin:
Added the description. Thanks guys!

PR #146, opened 2016-11-25
Adds Django REST Framework 3.5.x compatibility

### LINKED ISSUE #143: Add DRF 3.5 support

kciredor:
Fixes #143

### DIFF
--- a/dynamic_rest/filters.py
+++ b/dynamic_rest/filters.py
@@ -633,7 +633,7 @@ class DynamicSortingFilter(OrderingFilter):
 
         return valid_orderings, invalid_orderings
 
-    def get_valid_fields(self, queryset, view):
+    def get_valid_fields(self, queryset, view, context={}):
         """Return valid fields for ordering.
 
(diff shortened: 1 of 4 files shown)


dynamic_rest/filters.py
```diff
@@ -665,4 +669,8 @@ def get_valid_fields(self, queryset, view):
                 if not getattr(field, 'write_only', False) and
                 not field.source == '*' and field_name in valid_fields
             ]
+
+        if 'get_template_context' == caller:
```
aleontiev: This seems hacky / brittle -- why should the caller change the return value? Is there any other way we can retain compatibility that is caller-agnostic?

kciredor: I agree, if your PR #138 would finally get merged :-) then I could drop this hack entirely!

aleontiev: #140 is merged

kciredor:
Nice one @aleontiev - I have updated my PR accordingly. Ready to go :+1:

aleontiev:
Awesome, looks good

PR #147, opened 2016-11-28
attempt to fix 1.10 prefetch a__b nested issue

aleontiev:
During our Django upgrade work, we noticed that nested prefetches which worked in Django 1.9/1.8/1.7 are now misbehaving in Django 1.10.

This only applies to nested prefetch objects that use a queryset which itself prefetches a distant relationship using the original `a__b` prefetch syntax.

For example, the following query fails in Django 1.10 but works in earlier versions:

```
A.objects.all().prefetch_related(
    Prefetch(
        'b',
        queryset=B.objects.all().prefetch_related(
            'c__d'
        )
    )
)
# raises AttributeError: Cannot find 'd' on RelatedManager object, 'b__c__d' is an invalid parameter to prefetch_related()
```

However, if we change the internal prefetch of `c__d` to use the fully-nested syntax, all is well in Django 1.10:

```
A.objects.all().prefetch_related(
    Prefetch(
        'b',
        queryset=B.objects.all().prefetch_related(
            Prefetch(
                'c',
                queryset=C.objects.all().prefetch_related(
                    'd'
                )
            )
        )
    )
)
# No problem
```

TODOs:
- [x] Report this to Django: https://code.djangoproject.com/ticket/27554

### DIFF
--- a/dynamic_rest/filters.py
+++ b/dynamic_rest/filters.py
@@ -134,7 +134,7 @@ class FilterNode(object):
 
 
 class DynamicFilterBackend(BaseFilterBackend):
-    """A DRF filter backend that construct DREST querysets.
+    """A DRF filter backend that constructs DREST querysets.
 
     This backend is responsible for interpretting and applying
     filters, includes, and excludes to the base queryset of a view.
@@ -347,21 +347,37 @@ class DynamicFilterBackend(BaseFilterBackend):
 
     def _add_internal_prefetches(
         self,
+        model,
         prefetches,
         requirements
     ):
         """Add internal (required) prefetches to a prefetch dictionary."""
-        paths = requirements.get_paths()
-        for path in paths:
-            # Remove last segment, which indicates a field name or wildcard.
-            # For example, {model_a : {model_b : {field_c}}
-            # should be prefetched as a__b
-            prefetch_path = path[:-1]
-            key = '__'.join(prefetch_path)
-            if key:
-                prefetches[key] = key
-                if self.DEBUG:
-                    self._prefetches[key] = {}
+
+        for source, remainder in six.iteritems(requirements):
+            if not remainder or isinstance(remainder, six.string_types):
+                # no further requirements to prefetch
+                continue
+
+            related_field = get_model_field(model, source)
+            related_model = related_field.related_model
+
+            if related_model is None:
+                # GenericForeignKey has no related_model
+                # in this case, prefetch the generic relationship itself
+                prefetches[source] = source
+                continue
+
+            queryset = self._filter_queryset(
+                serializer=None,
+                model=related_model,
+                filters={},
+                queryset=None,
+                requirements=remainder
+            )
+            prefetches[source] = Prefetch(
+                source,
+                queryset=queryset
+            )
 
     def _add_request_prefetches(
         self,
@@ -440,6 +456,7 @@ class DynamicFilterBackend(BaseFilterBackend):
 
     def _filter_queryset(
         self,
+        model=None,
         serializer=None,
         filters=None,
         queryset=None,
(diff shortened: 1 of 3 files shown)


aleontiev:
@ryochiji PTAL -- there's some non-trivial changes to our prefetching logic being introduced here, i.e. all prefetch calls will now use nested Prefetch objects, whether they are driven by request parameters or by internal `requires` directives. This seems to be preferred given that we have more control when we use nested Prefetch objects and that is plays nice with Django. However, I wanted to loop you in early in case you think there's something crazy about that.

@benbardin FYI I was able to verify that these changes fix the issues we noticed downstream

aleontiev:
tox fails on 1.7, `related_field.related_model` needs a compatibility wrapper to work with old meta API

ryochiji:
In general, doing fully nested prefetches seems fine. I'll take a closer look at the code as well.

dynamic_rest/filters.py
```diff
@@ -460,96 +493,109 @@ def _filter_queryset(
         """
 
         is_root_level = False
-        if serializer:
-            if queryset is None:
-                queryset = serializer.Meta.model.objects
+        if model:
```
ryochiji: It looks like most of the code in this function is now wrapped in `if serializer`. Would it make sense to copy just the code that's relevant for the `if model` case in a separate method?

dynamic_rest/filters.py
```diff
@@ -347,21 +347,50 @@ def _build_prefetch_queryset(
...
+                # no further requirements to prefetch
+                continue
+
+            related_field = get_model_field(model, source)
+            related_model = related_field.related_model
+
+            if related_model is None:
+                # GenericForeignKey has no related_model
+                # in this case, prefetch the generic relationship itself
+                prefetches[source] = source
```
aleontiev: @ryochiji is this consistent with your understanding of prefetching + generic relationships? It made the query-counting tests pass and makes sense to me, but I'm not as familiar with the GFK stuff.

aleontiev:
@ryochiji please take another look. Refactored the queryset builders into 4 methods:

- `_build_queryset`: (entrypoint) builds a queryset that takes into account both internal and external reqs
- `_build_external_prefetches`: builds prefetches that take into account external reqs only
- `_build_internal_queryset`: builds a queryset that takes into account internal reqs only
- `_build_internal_prefetches`: builds prefetches that take into account internal reqs only 

I suspect we can further consolidate, but that seems like overkill at this point.

dynamic_rest/filters.py
```diff
@@ -306,72 +306,60 @@ def rewrite_filters(filters, serializer):
                 q &= ~Q(**{k: v})
         return q
 
-    def _build_prefetch_queryset(
+    def _build_internal_prefetches(
```
ryochiji: I wonder if we can be clearer than `internal` vs `external`. Maybe `implicit` (`inferred`?) vs `requested`?

aleontiev: :+1: I like those words

dynamic_rest/filters.py
```diff
@@ -186,21 +192,15 @@ def filter_queryset(self, request, queryset, view):
...
 
-        if self.DEBUG:
-            # in DEBUG mode, save a representation of the prefetch tree
-            # on the viewset
-            self.view._prefetches = self._prefetches = {}
-
-        return self._filter_queryset(queryset=queryset)
-
-    def _extract_filters(self, **kwargs):
+    def _get_filters(self, **kwargs):
```
ryochiji: maybe `_get_requested_filters`?

ryochiji:
General structure looks good. Looks like failed tests were against `_extract_filters` which got renamed (my bad for writing tests against a "private" method :-P).

aleontiev:
Fixed tests, going to merge

PR #148, opened 2016-12-15
add Python DREST client [WIP]

aleontiev:
TODO:
- [ ] Tests
- [ ] Documentation
- [ ] Update README

### DIFF
--- a/dynamic_rest/client.py
+++ b/dynamic_rest/client.py
@@ -21,7 +21,8 @@ class APIClient(object):
         username=None,
         password=None,
         token=None,
-        authorization_type='JWT'
+        authorization_type='JWT',
+        scheme='https'
     ):
         self.host = host
         self.version = version
@@ -34,6 +35,7 @@ class APIClient(object):
         self.username = username
         self.password = password
         self.token = token
+        self.scheme = scheme
         self.authorization_type = authorization_type
 
     @property
@@ -98,7 +100,7 @@ class APIClient(object):
                 prefix = '/%s' % prefix
 
             url = '%s%s' % (prefix, url)
-        return 'https://%s%s' % (self.host, url)
+        return '%s://%s%s' % (self.scheme, self.host, url)
 
     def save(
         self,
--- a/dynamic_rest/client.py
+++ b/dynamic_rest/client.py
@@ -1,0 +1,194 @@
+import json
+import requests
+import inflection
+
+
+API_AUTH_ENDPOINT = '/accounts/login/'
+
+
+class AuthenticationFailed(Exception):
+    pass
+
+
+class APIClient(object):
+
+    def __init__(
+        self,
+        host,
+        version=None,
+        session=None,
+        sessionid=None,
+        username=None,
+        password=None,
+        token=None,
+        authorization_type='JWT'
+    ):
+        self.host = host
+        self.version = version
+        self.session = session or requests.session()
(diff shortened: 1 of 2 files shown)


aleontiev:
@ryochiji FYI we have a few different implementations of this floating around so I wanted to centralize a standard solution into DREST. This is the most recent version I could find (from deploy-api)

dynamic_rest/client.py
```diff
@@ -0,0 +1,196 @@
...
+        key = resource if many else inflection.singularize(resource)
+        return body[key]
+
+    def get_resource_url(self, resource, pk=None):
+        return '%s/%s' % (
+            resource,
+            '%s/' % pk if pk else ''
+        )
+
+    def find(
```
aleontiev: @ryochiji thoughts on the interface here? I initially went with something simple: `find` (via `include` and `filters` params) and `save` with a data object. 

Long-term, I think it would be pretty cool if we could provide a Django interface for a DREST API, given how closely related they are... but that's a lot more work.

Regardless, curious to hear your thoughts. This could become very useful in a SOA where our services communicate with each other using DREST (we are doing exactly this between vishnu-api and a new service)

dynamic_rest/client.py
```diff
@@ -0,0 +1,194 @@
...
+        return self._token
+
+    @token.setter
+    def token(self, value):
+        self._token = value
+        self.authenticated = bool(value)
+        if value:
+            self.session.headers.update({
+                'Authorization': '%s %s' % (
+                    self.AUTHORIZATION_TYPE,
```
notsoluckycharm: Shouldn't this be lower cased to authorization_type?

aleontiev: :+1:

dynamic_rest/client.py
ryochiji: Hm... This might be more work, but how do you feel about a `QuerySet` like interface? Something that would look like:

```python
Resource.filter(foo=bar).include('field1', 'field2')
```

aleontiev: Yup, that's what I meant by a Django interface (QuerySet-like) -- I like it but I think its going to be a lot more work to get full support for this kind of interface.

ryochiji: Yeah, makes sense.

dynamic_rest/client.py
```diff
@@ -0,0 +1,196 @@
...
+    def find(
+        self,
+        resource,
+        *args,
+        **kwargs
+    ):
+        pk = args[0] if len(args) else None
+        many = False if pk else True
+        url = self.get_resource_url(resource, pk)
+        filters = kwargs.get('filters')
```
ryochiji: Any particular reason these are pulled from `kwargs` instead of being explicitly named params? Also, maybe add support for `exclude[]`, `sort`, and pagination?

dynamic_rest/client.py
```diff
@@ -0,0 +1,196 @@
...
+        include = kwargs.get('include')
+        params = []
```
ryochiji: Might be useful to just have a way to inject additional params (e.g. custom query params).

aleontiev: :+1:

aleontiev:
Closed in favor of https://github.com/AltSchool/dynamic-rest/pull/151

PR #151, opened 2017-01-18
Feature/api client edge

aleontiev:
@ryochiji please take a look, new take on Python client interface.

Sample usage:

```
+  Getting a client:
 +
 +        client = DRESTClient('my.api.io', authentication={'token': 'secret'})
 +
 +    Working with a resource:
 +
 +        User = client.users
 +
 +    Getting a single resource:
 +
 +        User.get('123')
 +
 +    Getting all resources (auto-paging):
 +
 +        User.all()
 +
 +    Getting filtered resources:
 +
 +        User.filter(name__icontains='john')
 +        other_users = User.exclude(name__icontains='john')
 +
 +    Including / excluding fields:
 +
 +        users = User.all()
 +        .excluding('birthday')
 +        .including('events.*')
 +        .get('123')
 +
 +    Mapping by field:
 +
 +        users_by_id = User.map()
 +        users_by_name = User.map('name')
 +
 +    Ordering results:
 +
 +        users = User.order_by('-name')
 +
 +    Updating records:
 +
 +        user = User.first()
 +        user.name = 'john'
 +        user.save()
 +
 +    Creating resources:
 +
 +        user = User.create(name='john')
```


dynamic_rest/fields/fields.py
```diff
@@ -97,7 +99,9 @@ def __init__(
         self._serializer_class = serializer_class
         self.bound = False
         self.queryset = queryset
-        self.embed = embed
+        self.sideloading = sideloading
```
aleontiev: Added this parameter to implementation, allowing consumers to force embedding/sideloading at a request level.

dynamic_rest/fields/fields.py
```diff
@@ -97,7 +99,9 @@ def __init__(
...
+        self.debug = debug
```
aleontiev: Added this parameter to allow consumers to request debug mode to be turned on, resulting in extra response metadata being returned.

dynamic_rest/serializers.py
```diff
@@ -66,19 +66,16 @@ def id_only(self):
...
-                    SideloadingProcessor(
-                        self,
-                        data
-                    ).data,
-                    serializer=self
-                )
-            else:
-                self._sideloaded_data = ReturnList(data, serializer=self)
-        return self._sideloaded_data
+            self._processed_data = ReturnDict(
```
aleontiev: Technically a breaking change, since this means that we will no longer return "data" without an envelope if a viewset requests that it should be returned without an envelope, if a viewset had sideloading disabled entirely.

dynamic_rest/processors.py
```diff
@@ -80,9 +80,14 @@ def process(self, obj, parent=None, parent_key=None, depth=0):
                 if not dynamic or getattr(obj, 'embed', False):
                     return
 
-                name = obj.serializer.get_plural_name()
-                pk = obj.pk_value or obj.instance.pk
+                serializer = obj.serializer
+                name = serializer.get_plural_name()
+                instance = getattr(obj, 'instance', serializer.instance)
 
+                if not instance:
```
ryochiji: This seems like a potentially breaking change since the previous code didn't require an instance?

aleontiev: Good catch, will change this back

dynamic_rest/serializers.py
ryochiji: What are the ramifications of this change?

ryochiji:
It looks like one of the things you're trying to accomplish is to add a mode where related objects come back nested. I actually did that in one of my hacks, but did it by modifying [`SideloadingProcessor.is_dynamic`](https://github.com/AltSchool/dynamic-rest/blob/hack/descriptive-fks/dynamic_rest/processors.py#L54). That approach seemed pretty un-intrusive but not sure if it's applicable to your scenario.

aleontiev:
@ryochiji I thought it would be useful to support an API parameter that can force embedding of all relationships, or force sideloading of all relationships. Default behavior is still to respect embeds at the serializer level.

dynamic_rest/serializers.py
aleontiev: This would breaks a viewset that defines `sideload = False` and expects serialized data to be returned without an envelope. In effect, this is standard DRF behavior, so I'm not sure why you would be operating in this mode.

dynamic_rest/processors.py
aleontiev: @ryochiji addressed this

dynamic_rest/serializers.py
ryochiji: 👍

dynamic_rest/viewsets.py
```diff
@@ -232,6 +232,26 @@ def get_request_fields(self):
...
+                debug = True
+            else:
+                debug = False
+        return debug
+
+    def get_request_sideloading(self):
+        sideloading = self.get_request_feature(self.SIDELOADING)
+        if sideloading:
+            sideloading = sideloading.lower()
+            if sideloading == 'true':
```
ryochiji: Might be nice to reuse [this list](https://github.com/AltSchool/dynamic-rest/blob/master/dynamic_rest/filters.py#L179) for consistent handling of boolean values.

dynamic_rest/client/client.py
```diff
@@ -0,0 +1,187 @@
...
+        other_users = client.users.exclude(name__icontains='john')
+
+    Including / excluding fields:
+
+        users = Users.all()
+        .excluding('birthday')
+        .including('events.*')
+        .get('123')
+
+    Mapping by field:
```
ryochiji: Nice!

dynamic_rest/client/client.py
```diff
@@ -0,0 +1,187 @@
...
+
+        users_by_id = Users.map()
+        users_by_name = Users.map('name')
+
+    Ordering results:
+
+        users = Users.order_by('-name')
```
ryochiji: The DREST param by default is `sort[]`... would it be better to be consistent with that? This is more consistent with Django though.

dynamic_rest/client/query.py
```diff
@@ -0,0 +1,168 @@
...
+        pages = meta.get('total_pages', 1)
+
+        self._data = self._load(data)
+        self._pages = pages
+        self._index = 0
+
+    def _load(self, data):
+        return self.resource.load(unpack(data))
+
+    def __iter__(self):
```
ryochiji: Probably not P1, but could be nice to also implement `__getitem__()` to support random access and slices :-)

dynamic_rest/client/client.py
aleontiev: I am open to either one, or both?

dynamic_rest/client/query.py
aleontiev: 👍 left a TODO for now. Definitely going to be a bit hairier, especially around caching (currently there is no cached data aside from current working page), so Im not going to take it on just yet

aleontiev:
@ryochiji any other concerns?

ryochiji:
Looks great! 👍

aleontiev:
@ryochiji I tried running vishnu-be tests against this PR and hit a few breaking tests due to the change in `.data`, e.g. when serializers are constructed manually by more complex viewsets. This was a rare case (1% of tests), but regardless, I decided to work a bit harder to retain backwards-compatibility.

To do this, I added a new serializer parameter called `envelope` which controls whether or not `.data` is wrapped in an envelope. The default is False, which provides the same behavior we have in DREST 1.5, but is set to True by the viewset. The `sideload` parameter (which controls both the envelope as well as deeper sideloading in 1.5) is still supported as a proxy for `envelope`, but is marked as pending deprecation.

Generally, `envelope` is preferred because it supports sideloading and response meta-data in an unambiguous way. However, when using serializers directly outside of the context of a view, I can see the value of stripping the envelope.

But how does that sound to you? :)

aleontiev:
Going ahead with this

dynamic_rest/serializers.py
```diff
@@ -213,11 +213,20 @@ def __init__(
...
         self.envelope = envelope
-        self.sideloading = sideloading
         self.debug = debug
         self.dynamic = dynamic
         self.request_fields = request_fields or {}
-        self.embed = embed if sideloading is None else not sideloading
+
+        # sideloading modifies top-level response keys,
+        # so it requires an envelope
+        if envelope is False:
```
ryochiji: Might be good to check for conflicting params, like `envelope == False and sideloading == True` and raise an error.

ryochiji:
👍 to the general approach. It might be nice to enforce some checks between `embed`, `envelope` and `sideloading` to avoid inconsistent configurations, but not urgent.

dynamic_rest/serializers.py
aleontiev: My initial approach here was fail-soft by overriding, e.g. similar to the way `include_fields='*'` overrides `exclude_fields`, rather than fail-hard by throwing an exception... but I'm open to being more strict

PR #153, opened 2017-02-21
DJ blueprints

aleontiev:
... and now, for something completely different: blueprints!

Integrating DREST into a new project and creating views/serializers is a multi-step process that can take a while for new developers. Furthermore, there is no established way to structure your API code, which leads to template copypasta and inconsistent implementations across projects.

This is by no means specific to DREST; I've noticed it to be a general shortcoming within the Python ecosystem and started the [dj](https://github.com/aleontiev/dj) project to help standardize code generation, among other common Python/Django developer workflows.

The change here introduces a `blueprints` folder with two blueprints: `init` and `api`. The `init` blueprint is run automatically when DREST is added to a project using `dj` and simply adds `dynamic_rest` and `rest_framework` to `settings.py:INSTALLED_APPS`. The `api` blueprint can then be used to generate versioned API endpoints, consisting of a serializer/viewset pair. These blueprints are integration-tested using a `dj` test helper.

For example, here's how you might use these blueprints to add an API to a brand new starter project (this mirrors the included integration test):

```
$ mkdir test-drest-dj
$ cd test-drest-dj
$ dj init dummy
...
$ dj generate model foo
...
$ dj migrate
...
$ dj add git+https://github.com/AltSchool/dynamic-rest.git@feature/dj-blueprints#egg=dynamic-rest
...
$ dj generate api v0 foo
...
$ dj serve
Running: manage.py runserver
Performing system checks...

System check identified no issues (0 silenced).
February 21, 2017 - 22:04:40
Django version 1.10, using settings 'dummy.settings'
Starting development server at http://127.0.0.1:8000/
Quit the server with CONTROL-C.
[21/Feb/2017 22:04:44] "GET /v0/foos/ HTTP/1.1" 200 11164 <-- our "foos" endpoint is live!
```


tests/integration/test_blueprints.py
```diff
@@ -0,0 +1,36 @@
+from unittest import TestCase
+from dj.test import TemporaryApplication
+from django.conf import settings
+import requests
+import time
+import json
+import os
+
+if settings.ENABLE_INTEGRATION_TESTS:
```
ryochiji: you can use the `@unittest.skipIf` decorator for this

ryochiji:
It's a neat idea. One idea I've had that somewhat overlaps with this, is adding more inferences to DREST so that you can just "register" a model and have it infer a default serializer / viewset on the fly. The way I was thinking about it, it should be possible to do it completely at run-time, rather than generating code, which might add some overhead, but has the advantage of not producing code that then potentially needs to be maintained.

I'm not opposed to this approach (and don't think they're necessarily mutually exclusive), but thought I'd put that out there. Also, `dj` sounds really cool. Definitely would be nice to get more consistency across projects and reduce boilerplate generation.

tests/integration/test_blueprints.py
aleontiev: :+1: will do, didn't know about that

aleontiev:
@ryochiji I do think we should pursue auto-registration / runtime endpoint generation as well. At the same time, there are cases where you would want to define an override for an auto-generated endpoint, e.g. views on ephemeral objects, serializers with custom method fields, views that need custom permissions, etc. I think we should aim to provide a reasonable baseline out-of-the-box that can be extended as necessary :)

PR #156, opened 2017-03-15
Allow setting Meta.list_serializer_class

fdintino:
The way that `DynamicModelSerializer` is currently written it is not possible to define a custom `list_serializer_class`, as that attribute is overridden in the `__new__` method. Provided that the user-defined `list_serializer_class` extends `DynamicListSerializer`, it should be settable.

### DIFF
--- a/dynamic_rest/serializers.py
+++ b/dynamic_rest/serializers.py
@@ -142,7 +142,11 @@ class WithDynamicSerializerMixin(WithResourceKeyMixin, DynamicSerializerBase):
         if not meta:
             meta = type('Meta', (), {})
             cls.Meta = meta
-        meta.list_serializer_class = DynamicListSerializer
+        list_serializer_class = getattr(
+            meta, 'list_serializer_class', DynamicListSerializer)
+        if not issubclass(list_serializer_class, DynamicListSerializer):
+            list_serializer_class = DynamicListSerializer
+        meta.list_serializer_class = list_serializer_class
         return super(
             WithDynamicSerializerMixin, cls
         ).__new__(


aleontiev:
:+1: looks good to me, thanks for your contribution :)

fdintino:
Thanks for the quick response!

harikvpy:
- dynamic-rest v2.3.0
- Django v4.2
- Python v3.11
- restframework v3.15

Has this feature been tested? I replaced the default `list_serializer_class`, first in the Serializer.Meta and then globally, but in both cases it doesn't seem to take effect. I would expect the `MyDynamicListSerializer.to_internal_value` to be called before the actual serializer's `to_internal_value`, but it doesn't seem to.

My code is pretty straightforward:

```
class InvoiceItemSerializer(DynamicModelSerializer):
    product = DynamicRelationField(ItemSerializer)

    class Meta:
        model = InvoiceItem
        fields = [
            'id', 'product', 'name', 'description', 'quantity', 'unit_price',
        ]
        read_only_fields = ['id']
        list_serializer_class = MyDynamicListSerializer

class InvoiceSerializer(DynamicModelSerializer):
    items = DynamicRelationField(InvoiceItemSerializer, many=True, deferred=True)

    class Meta:
        model = Invoice
        fields = [
            'id', 'date', 'number', 'customer', 'terms', 'notes', 'total', 'balance', 'items',
        ]
```

I even tried declaring:

```
class MyDynamicRelationField(DynamicRelationField):
    class Meta:
        list_serializer_class = MyDynamicListSerializer
```

and using it instead of DynamicRelationField, but to no avail.

Subsequently, I replaced the `list_serializer_class` in the global config:
```
class MyDynamicListSerializer(DynamicListSerializer):
        def __init__(self, *args, **kwargs):
            print('MyDynamicListSerializer.ctor')
            super().__init__(*args, **kwargs)

        def to_internal_value(self, data):
            print(f"MyDynamicListSerializer.to_internal_value - data: {data}")
            return super().to_internal_value(data)
```
```
DYNAMIC_REST = {
    list_serializer_class = 'utils.drf.MyDynamicListSerializer'.
}
```
This doesn't take effect either.

What could be going on?
@fdintino @aleontiev

PR #160, opened 2017-03-29
Lt/drest admin view

lynnetye:
@aleontiev , will make prettier!

screen shot 2017-03-29 at 12 28 10 pm

screen shot 2017-03-29 at 12 28 23 pm

screen shot 2017-03-28 at 4 35 51 pm


dynamic_rest/renderers.py
```diff
@@ -91,6 +91,9 @@ def process(result):
             c for c in context['columns'] if c not in self.COLUMN_BLACKLIST
         ]
         context['details'] = context['columns']
+        context['error_form'] = context['error_form']
```
aleontiev: This is a no-op

dynamic_rest/renderers.py
```diff
@@ -91,6 +91,9 @@ def process(result):
...
+        if context['error_form']:
```
aleontiev: I think this would break error-handling for single-record updates (edit/create one instance). To get around that, check for the presence of the "errors" key:

```
bulk_errors = context['response'].data.get('errors')
if context['error_form'] and bulk_errors:
   # this is a bulk upload error
   context['errors'] = bulk_errors
```

lynnetye: it shouldn't. 

it shouldn't! [see here](
https://github.com/AltSchool/dynamic-rest/pull/160/files#diff-c6db66dec7f23eb26c7417373b9160b0R244)

screen shot 2017-03-29 at 4 24 33 pm

aleontiev: Aah you're right. I mis-read the template logic, where you are already doing this check: `{% if error_form and not errors %}` corresponds to a single-upload error, and `{% if errors %}` correspond to bulk-upload error. :+1:

PR #161, opened 2017-03-30
Admin view + custom relationships [WIP]

aleontiev:


dynamic_rest/fields/generic.py
```diff
@@ -66,6 +64,7 @@ def get_pk_object(self, type_key, id_value):
     def to_representation(self, instance):
         try:
             # Find serializer for the instance
+            from dynamic_rest.routers import DynamicRouter
```
benbardin: why the inline import?

tests/serializers.py
```diff
@@ -96,40 +96,71 @@ class Meta:
...
         'UserSerializer',
         source='users',
         many=True,
         queryset=lambda srlzr: User.objects.filter(location_id=1),
         deferred=True)
 
+    # Custom getter/setter
+    loc1usersGetter = DynamicRelationField(
+        'UserSerializer',
+        source='*',
```
benbardin: Above, you write that `source='*'` means to `# ignore custom getter/setter`. Doesn't that mean this _isn't_ actually using the custom getter/setter, but rather that loc1users above is?

How does the role of `source` change after this PR?

tests/settings.py
```diff
@@ -36,7 +36,7 @@
     'django.contrib.contenttypes',
     'django.contrib.auth',
     'django.contrib.sites',
-    'debug_toolbar',
+    # 'debug_toolbar',
```
benbardin: remove?

tests/serializers.py
aleontiev: Source = '*' generally means "we don't know what the source is, there's some overrides we are going to rely on instead", e.g. this already happens by default with `DynamicMethodField`

dynamic_rest/fields/generic.py
aleontiev: This is to avoid a circular import between DynamicRouter and fields

dynamic_rest/serializers.py
```diff
@@ -43,10 +45,20 @@ def __init__(self, *args, **kwargs):
...
+
     def to_representation(self, data):
         iterable = data.all() if isinstance(data, models.Manager) else data
         return [self.child.to_representation(item) for item in iterable]
 
+    def get_natural_key(self):
+        """Get the child's natural key."""
+        return self.child.get_natural_key()
+
+    def get_url(self, pk=None):
```
aleontiev: TODO(ant): change this, DRF uses the `get_XXX` convention to define method-fields. Should also consider renaming `get_model`, although that is already in master...

dynamic_rest/serializers.py
```diff
@@ -301,6 +317,29 @@ def get_name(cls):
...
+                self.get_resource_key()
+            )
+        if pk:
+            return '%s/%s/' % (url, pk)
+        return url
+
+    @classmethod
+    def get_natural_key(cls):
+        if not hasattr(cls.Meta, 'natural_key'):
+            return 'name'
```
aleontiev: TODO(ant): more reasonable fallback behavior, e.g. look for a unique field on the model, or look through model fields for something unique

dynamic_rest/serializers.py
```diff
@@ -301,6 +317,29 @@ def get_name(cls):
         return cls.Meta.name
 
     @classmethod
+    def get_url(self, pk=None):
```
aleontiev: @ryochiji this is a pivotal part of the PR as it allows relation-fields and serializers to easily reference their endpoint, even for non-canonical serializers.

But there are a couple of issues:
- This assumes that a serializer is going to be defined by a single viewset, which may not be the case. If a serializer is registered with multiple viewsets, the current behavior is to take the URL of the latest viewset (arbitrary)
- Potentially taxing on performance, although this method would only be used for self-links (disabled by default) and the admin view, so don't think this would cause regressions for the JSON API. Can also probably fold this into the serializer cache

dynamic_rest/serializers.py
```diff
@@ -535,9 +574,11 @@ def to_representation(self, instance):
                     self
                 ).to_representation(instance)
 
-            if settings.ENABLE_LINKS:
-                # TODO: Make this function configurable to support other
-                #       formats like JSON API link objects.
+            query_params = self.get_request_attribute('query_params', {})
+            if (
+                settings.ENABLE_LINKS and
+                'exclude_links' not in query_params
```
aleontiev: Drive-by, noticed that the "exclude_links" parameter was not being respected

dynamic_rest/static/images/logo-white.svg
```diff
@@ -0,0 +1,15 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
aleontiev: TODO(ant): remove and replace with generic logo

dynamic_rest/templates/dynamic_rest/horizontal/select2.html
```diff
@@ -0,0 +1,80 @@
+{% load i18n %}
```
aleontiev: The only real template change: relationship fields based on select2 with query back to API

dynamic_rest/viewsets.py
```diff
@@ -1,24 +1,46 @@
...
 from dynamic_rest.pagination import DynamicPageNumberPagination
 from dynamic_rest.processors import SideloadingProcessor
-from dynamic_rest.renderers import DynamicBrowsableAPIRenderer
 from dynamic_rest.utils import is_truthy
 
 UPDATE_REQUEST_METHODS = ('PUT', 'PATCH', 'POST')
 DELETE_REQUEST_METHOD = 'DELETE'
 
 
+def get_view_name(view_cls, suffix=None):
```
aleontiev: TODO(ant): make this a view default

dynamic_rest/viewsets.py
```diff
@@ -448,7 +491,7 @@ def _create_many(self, data):
                 serializer.is_valid(raise_exception=True)
             except exceptions.ValidationError as e:
                 errors.append({
-                    'detail': str(e),
+                    'detail': e.detail,
```
aleontiev: Drive-by: `e.detail` is often a dictionary, `str(e)` returns a nasty string-encoded version of the dictionary

dynamic_rest/serializers.py
```diff
@@ -301,6 +317,29 @@ def get_name(cls):
...
+    def get_url(self, pk=None):
+        # if associated with a registered viewset, use its URL
+        url = getattr(self, '_url', None)
+        if url:
+            # use URL key to get endpoint
+            url = reverse(url)
+        if not url:
+            # otherwise, return canonical URL for this model
+            from dynamic_rest.routers import DynamicRouter
+            url = DynamicRouter.get_canonical_path(
```
ryochiji: Maybe this should have precedence over `reverse()`? Also, `get_canonical_path` can return `None`, so may also add a check for that below?

dynamic_rest/serializers.py
```diff
@@ -301,6 +317,29 @@ def get_name(cls):
         return cls.Meta.name
 
     @classmethod
+    def get_url(self, pk=None):
+        # if associated with a registered viewset, use its URL
+        url = getattr(self, '_url', None)
```
ryochiji: It looks like this is always being set in the router, so I think the canonical URL will never get used? Can we try to get the canonical path first and fall back on this if there isn't one?

dynamic_rest/serializers.py
aleontiev: :+1: to the safeguard, see my other comment about `get_canonical_path` vs `reverse`

dynamic_rest/serializers.py
ryochiji: Looking at the router code a little closer, I guess this _could_ be the canonical URL, but if there are multiple viewsets mapped to a single serializer, it could also be overwritten by something that's not canonical? Since canonical URLs are most useful when there is ambiguity, I think that might negate the purpose of canonical URLs and the explicit `register_resource` API.

aleontiev: This will only be set in the router if the serializer is associated directly with an exposed view. If the serializer has no view (e.g. a relationship that isn't exposed as "top-level" or a serializer that is used conditionally by a polymorphic view), then the canonical path (via serializer's model) will be used.

From the perspective of a serializer, I think its more intuitive that the URL is primarily derived from the routing hierarchy (API level), with a fallback to the underlying model context (ORM level). This is because the routing hierarchy will unambiguously resolve any connected serializer, whereas the model context will not.

This is particularly apparent in the (common) multi-version scenario. Lets say we have a V1 and a V2 representation of users, which means two distinct sets of urls/views/serializers that reference a common model. If we take V2 to be canonical after the release of V2, suddenly V1's serializer will start returning references to V2. If we take V1 to be canonical, we have the same issue in the other direction.

ryochiji: Got it. So this more like the "canonical URL for the serializer" as opposed to "canonical URL for the model". That makes sense 👍

aleontiev: Right; I think there's still value in the `register_resource` API in that it helps integration code outside of the API link arbitrary instances or models back to the API. 

However, I do think its worth examining the corner cases surfaced here and attempting to simplify. So far, here's what I'm hearing:

1) A single model can be represented by multiple serializers and endpoints, e.g. `/v1/users/` and `/v2/users/` 
2) A single serializer can be re-used across endpoints, e.g. `/v1/users/` and `/v1/people/` (different views of the same data)
3) A view can use different serializers depending on the request context, e.g. an admin version and a basic version (this is a DRF example)

This makes things very complicated in the worst case: models/serializers/views are all many-to-many. I think we should consider officially dropping support for 2) and 3) and arrive at e.g:

- A viewset has exactly one serializer and at most one model (ephemeral serializers have no model)
- A serializer has at most one viewset and at most one model (ephemeral serializers have no model, some serializers may not be exposed directly)
- A model can be represented by many serializers/viewsets

ryochiji: Yeah, dropping 2) doesn't seem too bad. If you have two similar serializers that need to share code, you can just extend a shared base class into two separate classes, so doesn't seem too burdensome. I don't think I've used 3) so don't have much context on why one would do that, but seems like the right solution should be to either differentiate behaviors in the serializer or just have separate viewsets. So, yeah, 👍 to dropping 2) and 3) from my end.

auvipy:
would be great if you pull and rebase against master to resolve conflicts

nnegri:
@ryochiji  @aleontiev We want to perform frontend regression testing on the [feature deploy](https://deploy.altschool.com/stacks/40407) for the admin product, so I rebased from master to take into account the new `ufd` login redirect change from last week (Zoe needs a reliable login flow for some of her tests). I'm now getting a [circle failure](https://circleci.com/gh/AltSchool/vishnu-backend/31627) for `dynamic_rest/test.py` on `instance.save()` in `create_instance`. Is this something you can address from here, or do you recommend I work around this somehow?

auvipy:
when we can try the 2.0 version? did you add drf 3.7 compat for 2.0 release?

auvipy:
sorry for bugging! can anyone give any idea about this work? will this ever be merged? what helps needed to complete this?

auvipy:
@aleontiev do you have plan to complete this? or it would be ok to take over?

aleontiev:
@auvipy, I've left the org and have since forked the project [here](https://github.com/asaak-co/dynamic-rest), where I have continued most of this work. Would love to merge in to the main project at some point, time permitting :)

auvipy:
@aleontiev I would love to help in maintaining this package :) you could add me as a co maintainer :)

PR #164, opened 2017-04-06
Ensure models aren't deleted multiple times

DeviateFish:
Given that `get_queryset` returns a configurable queryset, we may end up with an edge case where the same model is returned multiple times.

Ensure that the set is distinct before attempting to delete models, to prevent issues around trying to delete a model multiple times.


aleontiev:
:+1: feel free to merge after tests pass. Change looks good!

PR #165, opened 2017-04-10
Lt/drest admin view design

lynnetye:
@aleontiev , I don't trust my design skills that much but here is what it looks like incorporating our color palette. 

Desktop example:
screen shot 2017-04-10 at 11 28 58 am

Mobile example:
screen shot 2017-04-10 at 11 28 47 am


aleontiev:
@lynnetye nice!! looks great, much more approachable than the stock styling :)

PR #166, opened 2017-04-12
Upgrade/2017 04

atzannes:
I tried to run through the tutorial and I had lots of trouble because of legacy support and lack of updated support. This is documented in a detailed way in the commit messages, the highlights are:
* Django 1.7 is deprecated so I removed support for it from tox and docs
* django-debug-toolbar 1.4 incompatible with current sqlparse version (1.4 requires 0.1.x, need >=1.5 for 0.2x)
* BONUS: Django 1.11 is out! Added test configs to tox for Django 1.11 as well as REST 3.6
* Moved default location for db.sqlite3 (in the tutorial) so that it is not in a location that requires superuser privileges on linux

### DIFF
--- a/README.md
+++ b/README.md
@@ -63,8 +63,8 @@ but it can be used to provide fast and flexible CRUD operations to any consumer 
 # Requirements
 
 * Python (2.7, 3.3, 3.4, 3.5)
-* Django (1.7, 1.8, 1.9, 1.10)
-* Django REST Framework (3.1, 3.2, 3.3, 3.4, 3.5)
+* Django (1.8, 1.9, 1.10, 1.11)
+* Django REST Framework (3.1, 3.2, 3.3, 3.4, 3.5, 3.6)
 
 # Installation
 
@@ -609,10 +609,6 @@ Not all versions of Python, Django, and DRF are compatible. Here are the combina
 
 | Python | Django | DRF | OK  |
 | ------ | ------ | --- | --- |
-| 2.7    | 1.7    | 3.1 | YES |
-| 2.7    | 1.7    | 3.2 | YES |
-| 2.7    | 1.7    | 3.3 | YES |
-| 2.7    | 1.7    | 3.4 | NO<sup>4</sup> |
 | 2.7    | 1.8    | 3.1 | YES |
 | 2.7    | 1.8    | 3.2 | YES |
 | 2.7    | 1.8    | 3.3 | YES |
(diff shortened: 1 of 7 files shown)


atzannes:
Hm, I was under the impression that cicle had cleared this before issuing the PR. Let me work on it a bit to fix the issues.

README.md
```diff
@@ -665,17 +656,20 @@ Not all versions of Python, Django, and DRF are compatible. Here are the combina
...
+| 3.5    | 1.11   | 3.6 | YES |
 
 * 1: Django 1.9 is not compatible with DRF 3.1
 * 2: Django 1.9 is not compatible with Python 3.3
-* 3: Django 1.7 is not compatible with Python 3.5
-* 4: Django 1.7 is not compatible with DRF 3.4
-* 5: Django 1.10 is only compatible with DRF 3.4+
-
+* 3: Django 1.10 is only compatible with DRF 3.4+
+* 4: Django 1.10 is requires Python 2.7, 3.4, 3.5
```
aleontiev: typo: should be "Django 1.10 requires", no "is"

README.md
```diff
@@ -665,17 +656,20 @@ Not all versions of Python, Django, and DRF are compatible. Here are the combina
...
+* 5: Django 1.11 is requires Python 2.7, 3.4, 3.5, 3.6
```
aleontiev: typo: should be "Django 1.11 requires", no "is"

install_requires.txt
```diff
@@ -1,4 +1,4 @@
-Django>=1.7,<1.11
-djangorestframework>=3.1.0,<3.6.0
+Django>=1.8,<=1.11
+djangorestframework>=3.1.0,<3.6.2
```
aleontiev: Tox tests 3.6.2 specifically, this should be `<=3.6.2`

tests/settings.py
```diff
@@ -21,14 +21,16 @@
     # local sqlite database file
     DATABASES['default'] = {
         'ENGINE': 'django.db.backends.sqlite3',
-        'NAME': INSTALL_DIR + 'db.sqlite3',
+        'NAME': os.path.join(BASE_DIR, 'db.sqlite3'),
```
aleontiev: What is the reasoning behind this change, i.e. moving test database out of the install directory and into the tests directory?

aleontiev: Ah, I just read your comment re: Linux installation. If you use `pyenv` or `virtualenv`, I don't think this would be a problem, but lets go with this for increased robustness

atzannes: Just to be clear, the problem is the INSTALL_DIR in tests/settings.py (not the one in the Makefile):
tests/settings.py:INSTALL_DIR = '/usr/local/altschool/dynamic-rest/'

atzannes:
Thanks for catching those typos, they are fixed now. Are we ready to merge? (And do you prefer 'merge' or 'rebase-and-merge'?)

aleontiev:
Merge is fine!

atzannes:
:+1:  Thanks

atzannes:
we also need a tag v1.6.5 in git

PR #167, opened 2017-04-20
Add a method for getting the serializer from an instance, so that

rgangopadhya:
clients can override this if desired (to mock a generic relation, for
example)

Because we're pushing everything to `StudentCard`, in v4 API we still want `v4/evidences/?include[]=content_object.` to sideload using the correct serializer. In vishnu-backend we override the method provided here to accomplish that. https://github.com/AltSchool/vishnu-backend/pull/4046


ryochiji:
👍

PR #168, opened 2017-05-08
Fixed sphinx's python path

shanx:
I don't know about the setup you guys are using, but if I clone a fresh copy of dynamic-rest and run `make docs` you get a `ModuleNotFoundError: No module named 'tests.settings'`. This PR fixes that by adding the relative path to the dynamic-rest project to sys.path in sphinx's `conf.py`

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -15,6 +15,9 @@
 import os
 import sys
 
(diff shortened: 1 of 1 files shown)


aleontiev:
:+1: can reproduce the issue and the fix

PR #169, opened 2017-05-08
Added lookup_field to allow following relations using some slug instead of an id

shanx:
The use case for this feature is to allow following relations using descriptive URL's that contain a slug instead of just the pk. So instead of `http://localhost:9002/events/5/users/` allow API consumers to use `http://localhost:9002/events/new-event/users/`. This makes an API more descriptive.

This feature has quite some implications, I don't expect this PR to be accepted at once. But I think that at this point it is good to show this work to you guys as the original authors for feedback.


aleontiev:
@shanx looks interesting and I generally like the idea, but will take a closer look in the next few days. This definitely isn't a trivial change, so I'd want to make sure all of the corner-cases are addressed.

shanx:
Totally understand that, probably some things I'm doing in there you have better ways for as authors. If you have any questions, let me know. Thanks for considering this!

docs/tutorial.rst
```diff
@@ -454,5 +454,46 @@ To address this issue, DREST fields like DynamicField and DynamicMethodField sup
                 user.profile.preferred_last_name
             )
 
+Following relations using a slug instead of an id
+-------------------------------------------------
 
+DREST auto-generates an endpoint to return just the related data (useful as link objects). By default
```
aleontiev: Does this feature only impact related/link endpoints, or would it apply to the detail endpoint as well? In other words, would `localhost:9002/events/new-event/` work as well as `localhost:9002/events/5/`?

 It seems like the latter, but this documentation suggests it is specifically for links.

dynamic_rest/fields/fields.py
```diff
@@ -275,9 +275,17 @@ def to_representation(self, instance):
         serializer = self.serializer
         model = serializer.get_model()
         source = self.source
-        if not self.kwargs['many'] and serializer.id_only():
+        try:
+            lookup_field = serializer.Meta.lookup_field
```
aleontiev: `lookup_field = getattr(serializer.Meta, 'lookup_field', None)` would be more concise

dynamic_rest/serializers.py
```diff
@@ -147,6 +149,18 @@ def __new__(cls, *args, **kwargs):
...
         meta.list_serializer_class = list_serializer_class
+
+        # If a lookup field has been defined in the serializer that
+        # field becomes the ID field for the serializer
+        lookup_field = getattr(meta, 'lookup_field', None)
+        if lookup_field is not None:
+            model = getattr(cls.Meta, 'model')
+            info = model_meta.get_field_info(model)
+            # XXX: This might not be the way to correctly solve this
+            cls._declared_fields['id'] = \
```
aleontiev: This definitely seems a bit hacky; what is the significance of the field called "id"? Where in DRF/DREST is that used?

docs/tutorial.rst
shanx: will also apply to the detail endpoint, I'll update the docs!

dynamic_rest/serializers.py
shanx: It is done normally by metaclass https://github.com/encode/django-rest-framework/blob/a8e527a4625b48914a553256f8be2907bfcc7c99/rest_framework/serializers.py#L288 we want to replace the id field of the serializer runtime for this to work. I'll look again at this tomorrow with a fresh perspective, if in the meantime you have any additional thoughts on this.

dynamic_rest/fields/fields.py
shanx: The Meta attribute is not always available on the Serializer so the try/except is used for two levels instead of one

docs/tutorial.rst
shanx: done!

dynamic_rest/serializers.py
shanx: My bad! Replacing the id field on the serializer, what was I thinking? Referencing the resource is a separate thing from the contents of the resource all together, fixed in latest commit

shanx:
ping :)

shanx:
hey @aleontiev is there anything I can do still to help you guys get this merged in? Or are you guys just very busy at the moment? (which I can completely understand). cheers!

shanx:
@auvipy that conflict was a new one :) thanks for the heads up! just fixed it

MichaelCombs28:
Why hasn't this been merged yet?

shanx:
Dunno @MichaelCombs28, I think the guys are really busy. CC @aleontiev

Issue #152, opened 2017-02-13
explicite requirement of six


ornoone:
it seem you import six in ``utils.py``, but six is not in your ``install_requires.txt``.

PR #175, opened 2017-06-28
add six to dependencies to fix integration

aleontiev:
Addressing https://github.com/AltSchool/dynamic-rest/issues/152 and fixing integration test which triggers this issue


aleontiev:
@rgangopadhya I believe this will fix the issue you noticed

Issue #176, opened 2017-06-29
Sideloaded object not have all included fields


dpaluch-rp:
Hello I have a problem with sideloading, when I nest Serializer and include the same object multiple time, but with another fields not all fields are returned.

Location (pk 2) is included in Car (field: short_name) and Part (field: name) but in sideloaded object sometimes `short_name` is returned and sometimes `name` never both.

Example:

```
# models.py

class Location(models.Model):
    name = models.CharField(max_length=60)
    short_name = models.CharField(max_length=30)
    description = models.TextField()

class Car(models.Model):
    name = models.CharField(max_length=60)
    location = models.ForeignKey(Location)


class Part(models.Model):
    car = models.ForeignKey(Car)
    name = models.CharField(max_length=60)
    location = models.ForeignKey(Location)
```

```
# serializers.py

class LocationSerializer(DynamicModelSerializer):
    name = serializers.CharField(deferred=True)
    short_name = serializers.CharField(deferred=True)
    description = serializers.CharField(deferred=True)
    location = serializers.DynamicRelationField('LocationSerializer', many=True, deferred=True)


    class Meta:
        model = Location 

class PartSerializer(DynamicModelSerializer):
    name = serializers.CharField(deferred=True)
    description = serializers.CharField(deferred=True)
    location = serializers.DynamicRelationField('LocationSerializer', many=True, deferred=True)

    class Meta:
        model = Car

class CarSerializer(DynamicModelSerializer):
    parts = serializers.DynamicRelationField('PartSerializer', many=True, deferred=True)
    name = serializers.CharField(deferred=True)
    description = serializers.CharField(deferred=True)

    class Meta:
        model = Car
```

```
# viewsets.py

class CarViewset(DynamicModelViewset):
    serializer_class = CarSerializer

```

For query:

```localhost:8080/api/cars/1/?include[]=name&include[]=location.short_name&include[]=parts.name&include[]=parts.location.name```

I radomly get one of responses:

```
{
    "cars": [
        {
            "id": 1,
            "name": "Porshe",
            "location": 2
        }
    ],
    "locations": [
        {
            "id": 1
            "name": "China",
        },
        {
            "id": 2
            "short_name": "US",
        }
    ],
    "parts": [
      {
        "id": 2,
        "name": "wheel",
        "location": 2
      },
      {
        "id": 2,
        "name": "tire",
        "location": 1
      }
    ]
}
```

or

```
{
    "cars": [
        {
            "id": 1,
[...]

PR #187, opened 2017-07-14
#176 Fix: Sideloaded object not have all included fields

dpaluch-rp:
Fix problem with corresponding test case included


Dockerfile
```diff
@@ -0,0 +1,34 @@
+FROM ubuntu:xenial
```
aleontiev: Nice, thanks for the Dockerfile!

PR #191, opened 2017-08-08
Add import statement

PatDuJour:
Add import statement for WithDynamicModelSerializerMixin
The example for DREST-ifying Existing Serializer does not have an import statment. 
New developers with the package would have to lookup the module for WithDynamicModelSerializerMixin.


aleontiev:
:+1: nice catch!

PR #199, opened 2018-01-02
Allow view to inject additional filters in such a way that avoids this

rgangopadhya:
Django feature: https://code.djangoproject.com/ticket/18437

The issue is that these two querysets are not the same: 

`MasterCard.objects.filter(student_cards__is_archived=False).filter(student_cards__creator=<>)`
`MasterCard.objects.filter(student_cards__is_archived=False, student_cards__creator=<>)`

In the former, the `student_card` table is joined to `master_card` twice, and each join is filtered, whereas in the latter, `student_card` is joined once.

DREST takes all standard drest filters and wraps them into a single `filter` call, but adding custom filters outside of DREST ends up being a separate `filter` call, making it easy to run into this issue.

I've run into this a few times and thought it would make sense to allow callers a way to add filters into the DREST `filter` call, thus sidestepping this issue.

@ryochiji thought you might be the best candidate to review this. 

CC - @magistrula, this is my approach to the issue we ran into in this refactor https://github.com/AltSchool/vishnu-backend/pull/4744

### DIFF
--- a/dynamic_rest/filters.py
+++ b/dynamic_rest/filters.py
@@ -184,8 +184,18 @@ class DynamicFilterBackend(BaseFilterBackend):
         self.request = request
         self.view = view
 
+        # enable addition of extra filters (i.e., a Q())
+        # so custom filters can be added to the queryset without
+        # running into https://code.djangoproject.com/ticket/18437
+        # which, without this, would mean that filters added to the queryset
+        # after this is called may not behave as expected
+        extra_filters = getattr(self.view, 'extra_drest_filters', None)
+
         self.DEBUG = settings.DEBUG
-        return self._build_queryset(queryset=queryset)
+        return self._build_queryset(
+            queryset=queryset,
+            extra_filters=extra_filters,
+        )
 
     """
     This function was renamed and broke downstream dependencies that haven't
@@ -443,7 +453,8 @@ class DynamicFilterBackend(BaseFilterBackend):
         serializer=None,
         filters=None,
         queryset=None,
-        requirements=None
+        requirements=None,
+        extra_filters=None,
     ):
         """Build a queryset that pulls in all data required by this request.
 
@@ -530,6 +541,10 @@ class DynamicFilterBackend(BaseFilterBackend):
             excludes=filters.get('_exclude'),
             serializer=serializer
         )
+
+        # add additional filters specified by calling view
+        if extra_filters:
+            query = extra_filters if not query else extra_filters & query
 
         if query:
             # Convert internal django ValidationError to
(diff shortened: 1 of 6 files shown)


dynamic_rest/filters.py
```diff
@@ -184,8 +184,18 @@ def filter_queryset(self, request, queryset, view):
         self.request = request
         self.view = view
 
+        # enable addition of extra filters (i.e., a Q())
+        # so custom filters can be added to the queryset without
+        # running into https://code.djangoproject.com/ticket/18437
+        # which, without this, would mean that filters added to the queryset
+        # after this is called may not behave as expected
+        extra_filters = getattr(self.view, 'extra_drest_filters', None)
```
ryochiji: Also add to `WithDynamicViewSetMixin` and document.

ryochiji:
Please add test.

rgangopadhya:
@ryochiji added test

tests/viewsets.py
```diff
@@ -103,6 +104,27 @@ class LocationViewSet(DynamicModelViewSet):
...
+    serializer_class = LocationSerializer
+    queryset = Location.objects.all()
+
+    def filter_queryset(self, queryset):
+        user_name = self.request.query_params.get('user_name')
+        user_name_separate_filter = self.request.query_params.get(
+            'user_name_separate',
+        )
+        if user_name:
+            self.extra_drest_filters = Q(
```
ryochiji: Hm... I didn't realize you intended this to be dynamic. An attribute technically works, but given that the other DREST attributes are effectively constants, it would be more consistent to use a method: maybe something like `def get_extra_filters(self, request)`.

rgangopadhya:
Yeah, it needs to be dynamic to support custom filters, which are usually controlled by users -- see usage here https://github.com/AltSchool/vishnu-backend/pull/4744/files#diff-9e18d4e6bba0c20ab11c0a2ce60e370cR189. Good point, a method makes more sense -- modified.

dynamic_rest/viewsets.py
```diff
@@ -384,6 +384,14 @@ def list_related(self, request, pk=None, field_name=None):
 
         return Response(serializer.data)
 
+    def get_extra_filters(self, request):
+        # Override this method to enable addition of extra filters
+        # (i.e., a Q()) so custom filters can be added to the queryset without
+        # running into https://code.djangoproject.com/ticket/18437
+        # which, without this, would mean that filters added to the queryset
+        # after this is called may not behave as expected.
+        return None
```
rgangopadhya: Perhaps better to return `Q()` instead here and add the filters in every time. I think I was hesitant to do this because of this https://code.djangoproject.com/ticket/24279, which according to `git tag --contains afe0bb7b13bb8dc4370f32225238012c873b0ee3`, was only introduced into Django in 1.9.

ryochiji: I think returning None is fine.

PR #202, opened 2018-01-10
[FastQuery] Use proper pk attribute names

DeviateFish:
Use `pk.attname` instead of `pk.name`, since `attname` resolves to the actual column name (and hence the right key for the `dict` representation of the model)


notsoluckycharm:
LGTM 👍 tests seem to be failing on syntax which

PR #203, opened 2018-01-10
[FastQuery] Support for DynamicGenericRelationField

DeviateFish:
Adds compatibility with `DynamicGenericRelationField`, which provides no queryset to `Prefetch`


ryochiji:
Hm... are GFKs being prefetched? Concerned about N+1 situation here (where it might "work" but it's fetching required data one at a time).

DeviateFish:
I'll look into the N+1 on this, but I do believe they were being prefetched.  They just weren't being serialized properly.

PR #205, opened 2018-01-19
[FAST QUERY] Model Opt Out + Logic 

notsoluckycharm:


dynamic_rest/prefetch.py
```diff
@@ -205,13 +239,24 @@ def execute(self):
         # TODO: check if queryset already has values() called
         # TODO: use self.fields
         qs = self.queryset._clone()
-        data = list(qs.values())
 
-        self.merge_prefetch(data)
+        has_fastquery = hasattr(self.model, 'use_fastquery')
+        use_models = has_fastquery and self.model.use_fastquery == False
```
DeviateFish: Would probably abbreviate this to `getattr(self.model, 'use_fastquery', False)`

dynamic_rest/prefetch.py
```diff
@@ -41,6 +41,40 @@ def __getattr__(self, name):
...
+        return self[self.pk_field]
+
+    def __getitem__(self, value):
+
+        if hasattr(self.data, str(value)):
+            return getattr(self.data, str(value))
+
+        # for the purposse of mapping serialized model + '_id' fields back to
+        # internal models, we need to check if that pattern is present
+        is_nested_obj = value.split('_')
```
DeviateFish: Can I get some explanation why this ends up being needed?  I'm not sure I follow the usage here.

notsoluckycharm: I can tell you why I did it, and my theory is [here](https://github.com/AltSchool/dynamic-rest/blob/2f744d150015341a6e4d6eca6c266373703208bd/dynamic_rest/serializers.py#L648)

One of the attributes requested was, I think, preview_of_id from the parent serializer because it's a related model.

That attribute doesn't exist on this model and so to return something I look to see if it follows that format and return the value it asks for.

dynamic_rest/prefetch.py
notsoluckycharm: Are you sure? Every model in server.models does not define this variable so would the default be True then instead of False?

DeviateFish: Should be able to.  The last argument to `getattr` is the default to use if the attribute doesn't exist.  It's basically the same as what you have, or something like `self.model.use_fastquery if hasattr(self.model, 'use_fastquery') else False`

dynamic_rest/prefetch.py
DeviateFish: Ohh, interesting.  Yeah, this should work for now.  There's probably a more formal way to do it (using through fields or something), but this should work for now.  We'll run with it and see if any other things shake out.

dynamic_rest/prefetch.py
```diff
@@ -205,13 +239,24 @@ def execute(self):
...
 
-        self.merge_prefetch(data)
+        has_fastquery = hasattr(self.model, 'use_fastquery')
+        use_models = has_fastquery and self.model.use_fastquery == False
+
+        if use_models:
+            self._data = FastList(
+                map(lambda obj: SlowObject(
+                    obj, pk_field=self.pk_field
+                ), qs.all())
```
ryochiji: If there are downstream prefetches, this queryset doesn't have those. It might be possible to here get `self.prefetches`, and pass to `qs.prefetch_related()`.

dynamic_rest/prefetch.py
```diff
@@ -205,13 +239,32 @@ def execute(self):
         # TODO: check if queryset already has values() called
         # TODO: use self.fields
         qs = self.queryset._clone()
-        data = list(qs.values())
 
-        self.merge_prefetch(data)
+        has_fastquery = hasattr(self.model, 'use_fastquery')
```
DeviateFish: So, echoing Ryo's other comment, this should probably be something like `USE_FASTQUERY` to make it clear that it's a constant.

DeviateFish:
Looks good for now.  I have some more changes that need to be dropped on top of these, so I'm going to merge this and open a new PR

PR #206, opened 2018-01-19
Try fixing some prefetch stuffs

DeviateFish:
Make things more consistent across `get_attribute` and `to_representation`, to avoid madness where `instance` is sometimes an instance, and sometimes the related model.

### DIFF
--- a/dynamic_rest/prefetch.py
+++ b/dynamic_rest/prefetch.py
@@ -244,6 +244,14 @@ class FastQuery(FastQueryCompatMixin, object):
         use_models = has_fastquery and self.model.use_fastquery == False
 
         if use_models:
+            prefetches = [
+                Prefetch(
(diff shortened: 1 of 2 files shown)


dynamic_rest/prefetch.py
```diff
@@ -244,6 +244,14 @@ def execute(self):
...
 
         if use_models:
+            prefetches = [
+                Prefetch(
+                    prefetch.field,
+                    queryset=prefetch.query.queryset
+                ) for prefetch in self.prefetches.values()
+            ]
+            if len(prefetches) > 0:
+                qs = qs.prefetch_related(*prefetches)
```
ryochiji: This makes all downstream prefetches/sideloads also slow, but sounds like that's deliberate for now.

PR #207, opened 2018-01-22
[FastQuery] Fix Prefetches and field validation

DeviateFish:
Fixes prefetch logic on the slowquery side, and updates validation for related fields.

Also fixes a merge regression.

### DIFF
--- a/dynamic_rest/fields/fields.py
+++ b/dynamic_rest/fields/fields.py
@@ -270,26 +270,47 @@ class DynamicRelationField(WithRelationalFieldMixin, DynamicField):
     def get_attribute(self, instance):
         serializer = self.serializer
         model = serializer.get_model()
+
         # attempt to optimize by reading the related ID directly
         # from the current instance rather than from the related object
         if not self.kwargs['many'] and serializer.id_only():
             return instance
-        else:
+        elif model is not None:
             try:
                 return getattr(instance, self.source)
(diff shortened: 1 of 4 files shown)


dynamic_rest/fields/fields.py
```diff
@@ -269,40 +269,55 @@ def _is_dynamic(self):
...
+                # this is probably a one-to-one field, or a reverse related
+                # lookup, so let's look it up the slow way and let the
+                # serializer handle the id dereferencing
+                try:
+                    instance = getattr(instance, source)
+                except model.DoesNotExist:
+                    instance = None
+
+        # dereference ephemeral objects
+        if model is None:
```
ryochiji: this could just be moved to `get_attribute()` too

dynamic_rest/fields/fields.py
```diff
@@ -269,40 +269,55 @@ def _is_dynamic(self):
...
+            if hasattr(instance, source_id):
+                return getattr(instance, source_id)
+            elif model is not None and not use_fastquery:
+                # this is probably a one-to-one field, or a reverse related
+                # lookup, so let's look it up the slow way and let the
+                # serializer handle the id dereferencing
+                try:
+                    instance = getattr(instance, source)
+                except model.DoesNotExist:
+                    instance = None
```
ryochiji: probably could just `return None` here

PR #209, opened 2018-01-29
[Prototype] FastQuery

DeviateFish:
Adds functionality to support usage of DREST serializers with `.values` queries.

With large and complex requests, there comes significant overhead within django itself, primarily from marshaling large numbers of models.  The deeper or more complex the model graph, this worse this scales.

FastQuery attempts to bypass a large amount of that model serialization by using the `dict` representation returned by the database direction, by way of `queryset.values()`.  This is accomplished by building some api-compatible wrappers that work with the underlying dicts while keeping the internals of that abstracted away from layers beyond the serializer layer.

This isn't a perfect solution, or even likely to be the best one.  Given that it's a prototype and of dubious long-term worth, this is not something that will be enabled by default, and will take some work to actually enable.


dynamic_rest/prefetch.py
```diff
@@ -0,0 +1,482 @@
...
+
+        if hasattr(self.data, str(value)):
+            return getattr(self.data, str(value))
+
+        # for the purposse of mapping serialized model + '_id' fields back to
+        # internal models, we need to check if that pattern is present
+        is_nested_obj = value.split('_')
+        test_attr = '_'.join(is_nested_obj[:-1])
+        attr_exists = hasattr(self.data, test_attr)
+        if is_nested_obj[-1] == 'id' and attr_exists:
```
ryochiji: Kind of curious about this... in what cases would `*_id` not be available? Doesn't Django automatically provide those for all single-relations?

dynamic_rest/prefetch.py
```diff
@@ -0,0 +1,482 @@
...
+            queryset = queryset.all()
+        self.queryset = queryset
+        self.model = queryset.model
+        self.prefetches = {}
+        self.fields = None
+        self.pk_field = queryset.model._meta.pk.attname
+        self._data = None
+        self._my_ids = None
+
+    # TODO: probably override __iter__ or something
```
ryochiji: remove TODO since this appears to have been done

dynamic_rest/serializers.py
```diff
@@ -504,11 +568,44 @@ def _faster_to_representation(self, instance):
...
         for field in fields:
-            try:
-                attribute = field.get_attribute(instance)
-            except SkipField:
-                continue
+            # we exclude dynamic fields here because the proper fastquery
+            # dereferencing happens in the `get_attribute` method now
+            if (
+                is_fast and
+                not isinstance(
```
ryochiji: This check would happen for each field on each object instance being serialized. Probably not a huge overhead, but would be more optimal to pre-select the non-(DGRF,DRF) fields once at the serializer level, then loop over those fields.

dynamic_rest/serializers.py
```diff
@@ -504,11 +568,44 @@ def _faster_to_representation(self, instance):
...
+                    attribute = instance.get(field.source + '_id')
+                    ret[field.field_name] = attribute
+                    continue
+                else:
+                    try:
+                        attribute = instance[field.source]
+                    except KeyError:
+                        # slower, but does more stuff
+                        # Also, some temp debugging
+                        if hasattr(instance, field.source):
```
ryochiji: I think this check was for debugging. We could probably remove and just do `getattr(instance, field.source, None)`.

tests/test_serializers.py
```diff
@@ -707,11 +707,14 @@ def test_get_all_fields(self):
         home_field_1 = self.serializer.fields['home']
         home_field_2 = all_fields['home']
 
+        '''
+        # Expected with fields cache
```
ryochiji: maybe remove this block, or wrap in check to only run if fields caching is disabled.

ryochiji:
@DeviateFish - some comments, but no blockers if you want to address those later.

dynamic_rest/prefetch.py
DeviateFish: I'm guessing this would handle the case where the foreign key has a custom column name.

dynamic_rest/serializers.py
DeviateFish: I'll do this one as an optimization in a follow-up

dynamic_rest/serializers.py
DeviateFish: I'm going to leave this in for now.  While making some modifications to the FastQuery code, I definitely saw these warning messages from time to time... They were helpful in debugging, since they usually arose when something was actually wrong.

PR #222, opened 2018-04-11
Only patch prefetch_one_level for Django versions < 2.0.0

ben-prezola:


ben-prezola:
Already fixed. My bad

PR #224, opened 2018-04-17
prevents execution of base query through accessing query attribute on…

erikdbrown:
… prefetch object
Accessing the `query` property on the prefetch object was causing the execution of the base query.

For example, a call to: `/table/1`, these queries would execute:
1. `SELECT DISTINCT "foo"."property" FROM "bar"`
1. `SELECT DISTINCT "foo"."property" FROM "bar" WHERE "foo"."id" IN ('1')`

This change will prevent the first execution

The separation of lines was to fit within character limit

### DIFF
--- a/dynamic_rest/prefetch.py
+++ b/dynamic_rest/prefetch.py
@@ -227,7 +227,8 @@ class FastQueryCompatMixin(object):
 
         prefetches = []
         for field, fprefetch in self.prefetches.items():
-            qs = fprefetch.query.queryset if fprefetch.query else None
+            has_query = hasattr(fprefetch, 'query')
+            qs = fprefetch.query.queryset if has_query else None
             prefetches.append(
                 Prefetch(field, queryset=qs)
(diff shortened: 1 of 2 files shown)


ryochiji:
@erikdbrown - can you bump the version in `setup.py`?

erikdbrown:
bumped, @ryochiji

erikdbrown:
@ryochiji @DeviateFish I pushed up the new `v1.8.1` tag to `master` and attempted to test a publish according to `CONTRIBUTING.md` by running `make pypi_upload_test`. This failed after entering a password, so I likely don't have the correct credentials to publish.

ryochiji:
Yeah, no worries. I usually do the package releases.

ryochiji:
Also, for future reference, I usually use the [GitHub UI](https://github.com/AltSchool/dynamic-rest/releases) for releases since it makes it easier to add release notes.

Issue #232, opened 2018-06-21
Bulk update fails with UUID primary key


ben-prezola:
The lookup key in `DynamicListSerializer.update` in my case is a UUID so the data lookup fails.

PR #233, opened 2018-06-21
Update DynamicListSerializer

ben-prezola:
Cast lookup key to string to allow for UUID primary keys

Refs #232

### DIFF
--- a/dynamic_rest/serializers.py
+++ b/dynamic_rest/serializers.py
@@ -99,7 +99,7 @@ class DynamicListSerializer(
         lookup_attr = getattr(self.child.Meta, 'update_lookup_field', 'id')
 
         lookup_objects = {
-            entry.pop(lookup_attr): entry
+            str(entry.pop(lookup_attr)): entry
             for entry in validated_data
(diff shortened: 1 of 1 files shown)


ryochiji:
@ben-prezola - I merged this but it caused test failures in `BulkUpdateTestCase.test_bulk_update_fail_without_lookup_attribute` so I reverted it. If you can resolve that, I'd be happy to take a look again. Also note that the next release will contain a config setting to [override the list serializer](https://github.com/AltSchool/dynamic-rest/blob/master/dynamic_rest/conf.py#L36), so this could be done as a local override as well.

ben-prezola:
@ryochiji Sorry, I didn't test it. That's useful to know about the list serializer setting. I do think however that it is quite common to use UUIDs as primary keys and so overriding `update` would require a lot of duplication just to check for this.

I have created a new PR https://github.com/AltSchool/dynamic-rest/pull/243 that explicitly catches UUIDs rather than just casting everything to a string and can confirm the tests now pass.

PR #234, opened 2018-06-23
fix couple of inefficiencies in POST/PATCH flows

ryochiji:
Eliminate some unnecessary queries in the PATCH flow.

### DIFF
--- a/dynamic_rest/filters.py
+++ b/dynamic_rest/filters.py
@@ -192,11 +192,14 @@ class DynamicFilterBackend(BaseFilterBackend):
         # after this is called may not behave as expected
         extra_filters = self.view.get_extra_filters(request)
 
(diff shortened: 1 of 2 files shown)


dynamic_rest/filters.py
```diff
@@ -192,11 +192,14 @@ def filter_queryset(self, request, queryset, view):
         # after this is called may not behave as expected
         extra_filters = self.view.get_extra_filters(request)
 
+        disable_prefetches = self.view.is_update()
```
ryochiji: On PATCH requests, the queryset is executed twice (once to modify the object, then again after the modification) but the first time around not all prefetches are required (we do `include=*` to be safe but only fields actually being modified are needed). So rather than using prefetches, it ends up being more efficient to let the ORM fetch what it needs.

Issue #171, opened 2017-05-18
ordering direction


Incanus3:
Hi,
I know this isn't the right place for Q&A, but I haven't found any documented place to ask questions, so I'm asking here. Is there a way to dynamically sort records in descending order, e.g. something like `/clients?sort[]=name.desc&page=2&per_page=5` (this doesn't work obviously). I could sort the records on the client side, but that doesn't cover the pagination case.

aleontiev:
I'm not quite sure what you mean -- can you give me an example as a Django queryset / Python code? If its possible to translate your request into a single queryset, then we should be able to support it.

Incanus3:
When looking up the QuerySet documentation to achieve this, I found out you can add a minus sign before the column names given to `.order_by()` to sort in descending order and it seems to be working in the dynamic-rest sort param as well. Before, I was looking at the [dynamic-rest documentation](http://dynamic-rest.readthedocs.io/en/latest/), the DRF ViewSet documentation and event the DynamicSortingFilter code [here](https://github.com/AltSchool/dynamic-rest/blob/master/dynamic_rest/filters.py#L570) to find out if this was possible, but it didn't occur to me to look as deep as the Django QuerySet documentation. Sorry to waste your time and thank you for nudging me in the right direction. It might be nice to document this [here](https://github.com/AltSchool/dynamic-rest#ordering) though. Should I update the docs and create a PR?

aleontiev:
Ah I misunderstood -- we do support sorting by descending order, simply prefix your field with a `-` sign as you would with Django: for example, `/clients?sort[]=-name&page=2&per_page=5`.

I didn't realize that this was not documented in our README, feel free to create a PR to address this. Thanks!

Incanus3:
I sure will, but there's one more thing I'd like to address before updating the README. [Here](https://github.com/AltSchool/dynamic-rest#ordering) the README says that you can sort by fields of related records as well, but it doesn't seem to work. I traced the problem to [this condition](https://github.com/AltSchool/dynamic-rest/blob/master/dynamic_rest/filters.py#L638) which returns `false` for nested fields. The `valid_fields_map` only contains the direct serializer fields (including the `DynamicRelationField`s), but not the nested fields (using dot notation), so the aforementioned condition can't possibly return `true` for nested fields.

aleontiev:
[...]

PR #236, opened 2018-06-28
Add support for ordering by nested fields #171

joshvillbrandt:
Rebasing @Incanus3's commit from #171.

https://github.com/AltSchool/dynamic-rest/compare/master...Incanus3:feature/ordering-by-related-fields

https://github.com/AltSchool/dynamic-rest/compare/d564cd10437cbd136db68a514481da29fd2a5ae7...master#diff-fbc210b7daea111dceb4ac7cec923baf

### DIFF
--- a/dynamic_rest/filters.py
+++ b/dynamic_rest/filters.py
@@ -676,11 +676,6 @@ class DynamicSortingFilter(OrderingFilter):
         Overwrites the DRF default remove_invalid_fields method to return
         both the valid orderings and any invalid orderings.
         """
-        # get valid field names for sorting
-        valid_fields_map = {
-            name: source for name, source in self.get_valid_fields(
-                queryset, view)
-        }
 
         valid_orderings = []
         invalid_orderings = []
@@ -691,22 +686,58 @@ class DynamicSortingFilter(OrderingFilter):
             stripped_term = term.lstrip('-')
             # add back the '-' add the end if necessary
             reverse_sort_term = '' if len(stripped_term) is len(term) else '-'
-            if stripped_term in valid_fields_map:
-                name = reverse_sort_term + valid_fields_map[stripped_term]
-                valid_orderings.append(name)
+            ordering = self.ordering_for(stripped_term, view)
+
+            if ordering:
+                valid_orderings.append(reverse_sort_term + ordering)
             else:
                 invalid_orderings.append(term)
 
         return valid_orderings, invalid_orderings
 
-    def get_valid_fields(self, queryset, view, context={}):
-        """Return valid fields for ordering.
-
-        Overwrites DRF's get_valid_fields method so that valid_fields returns
-        serializer fields, not model fields.
+    def ordering_for(self, term, view):
         """
-        valid_fields = getattr(view, 'ordering_fields', self.ordering_fields)
+        Return ordering (model field chain) for term (serializer field chain)
+        or None if invalid
 
+        Raise ImproperlyConfigured if serializer_class not set on view
+        """
+        if not self._is_allowed_term(term, view):
+            return None
+
+        serializer = self._get_serializer_class(view)()
+        serializer_chain = term.split('.')
+
+        model_chain = []
+
+        for segment in serializer_chain[:-1]:
+            field = serializer.get_all_fields().get(segment)
+
+            if not (field and field.source != '*' and
+                    isinstance(field, DynamicRelationField)):
+                return None
+
+            model_chain.append(field.source or segment)
+
+            serializer = field.serializer_class()
+
+        last_segment = serializer_chain[-1]
+        last_field = serializer.get_all_fields().get(last_segment)
+
+        if not last_field or last_field.source == '*':
+            return None
+
+        model_chain.append(last_field.source or last_segment)
+
+        return '__'.join(model_chain)
+
+    def _is_allowed_term(self, term, view):
+        valid_fields = getattr(view, 'ordering_fields', self.ordering_fields)
+        all_fields_allowed = valid_fields is None or valid_fields == '__all__'
+
+        return all_fields_allowed or term in valid_fields
+
+    def _get_serializer_class(self, view):
         # prefer the overriding method
         if hasattr(view, 'get_serializer_class'):
             try:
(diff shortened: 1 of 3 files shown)


dynamic_rest/filters.py
```diff
@@ -691,22 +686,66 @@ def remove_invalid_fields(self, queryset, fields, view):
...
+        Raise ImproperlyConfigured if serializer_class not set on view
         """
+        if not self._is_allowed_term(term, view):
+            return None
+
+        serializer = self._get_serializer_class(view)()
+        serializer_chain = term.split('.')
+
+        if len(serializer_chain) == 1:
+            field = serializer.get_all_fields().get(term)
```
joshvillbrandt: I'm using `serializer.get_all_fields()` here instead of `serializer.fields` so that sorting works with deferred relationships as well.

dynamic_rest/filters.py
```diff
@@ -691,22 +686,66 @@ def remove_invalid_fields(self, queryset, fields, view):
...
-        serializer fields, not model fields.
+        Raise ImproperlyConfigured if serializer_class not set on view
         """
+        if not self._is_allowed_term(term, view):
+            return None
+
+        serializer = self._get_serializer_class(view)()
+        serializer_chain = term.split('.')
+
+        if len(serializer_chain) == 1:
```
ryochiji: I believe this `if ... else` block is unnecessary, and what's currently in the `else` block should work in all cases (if `serializer_chain` is size of 1, the for loop is simply skipped, but the rest should do the right thing).

dynamic_rest/filters.py
```diff
@@ -691,22 +686,66 @@ def remove_invalid_fields(self, queryset, fields, view):
...
+    def ordering_for(self, term, view):
+        """
+        Return ordering (model field chain) for term (serializer field chain)
+        or None if invalid
 
-        Overwrites DRF's get_valid_fields method so that valid_fields returns
-        serializer fields, not model fields.
+        Raise ImproperlyConfigured if serializer_class not set on view
         """
+        if not self._is_allowed_term(term, view):
```
ryochiji: Since we're checking `term` here, that means if `term` is nested we also expect that to appear in the view's `ordering_fields` attribute verbatim. Is that the intention? (i.e. if we want to sort by `foo.bar` or `foo.baz`, or `foo.bam`, `ordering_fields` needs to contain all of those, rather than, say, just `foo`.)

ryochiji:
@joshvillbrandt - Thanks for putting this PR together. Would you be able to add a quick test as well? It could be added alongside existing sort-related tests in `tests/test_api.py`.

joshvillbrandt:
Hey @ryochiji, thanks for the review! Yeah, I'll respond to the feedback and create a test tomorrow!

dynamic_rest/filters.py
joshvillbrandt: I didn't actually read through this originally since I based this off of @Incanus3's commit, but yeah, great point. The `if ... else` block has been removed and I've kept the `else` section.

dynamic_rest/filters.py
joshvillbrandt: I can't speak to the original intention, but I agree that this is a bit weird. I think what is weird about this is attempting to define ordering fields on the view as opposed to the model or the serializer.

So long as `ordering_fields` is supported on the view, I think I am fine with the current implementation. While more verbose, the current implementation affords the ability to block specific nested fields from ordering where just putting `foo` would not.

I am in favor of leaving this as-is for now. In the future, someone could support an addition `ordering_fields` parameter on the serializer or the model to allow for more sustainable ordering fields control.

Thoughts?

joshvillbrandt:
Okay @ryochiji, feedback has been addressed and two tests have been added. Let me know what you think!

setup.py
```diff
@@ -11,7 +11,9 @@
     description=DESCRIPTION,
     include_package_data=True,
     dependency_links=open('dependency_links.txt').readlines(),
-    install_requires=open('install_requires.txt').readlines() if sys.version_info.major == 2 else open('install_requires_python3.txt').readlines(),
+    install_requires=open('install_requires.txt'
+                          if sys.version_info.major == 2
+                          else 'install_requires_python3.txt').readlines(),
```
joshvillbrandt: This line was failing the flake8 test (is this repo tested via CI?!) so I had to fix it to be able to run the rest of the tests.

tests/test_api.py
```diff
@@ -1311,7 +1343,7 @@ class TestDogsAPI(APITestCase):
     def setUp(self):
         self.fixture = create_fixture()
 
-    def test_sort(self):
+    def test_sort_implied_all(self):
```
joshvillbrandt: I renamed a few of these. Happy to undo the name changes if you don't like them.

joshvillbrandt:
Ping @ryochiji !

setup.py
ryochiji: Our recommended wrapping style would be something like:

```python
    install_requires=open(
        'install_requires.txt' if sys.version_info.major == 2
        else 'install_requires_python3.txt'
    ).readlines(),

File: tests/test_api.py [hunk 6]
```diff
@@ -858,6 +858,38 @@ def test_boolean_filters_on_boolean_field(self):
...
+        with self.assertNumQueries(1):
+            response = self.client.get(url)
+        self.assertEquals(200, response.status_code)
+        self.assertEquals(
+            {
+                'users': [
+                    {'id': 4, 'location': 3, 'name': '3'},
+                    {'id': 3, 'location': 2, 'name': '2'},
+                    {'id': 1, 'location': 1, 'name': '0'},
+                    {'id': 2, 'location': 1, 'name': '1'},
```
ryochiji: Not sure how stable the last 2 items are and the potential for intermittent test failures. Maybe we can do something like:

```python
response = json.loads(response.content.decode('utf-8')))
self.assertEquals(
    [3, 2, 1, 1],
    [row['location'] for row in response['users']]
)
```

tests/test_api.py
ryochiji: Thanks for doing that!

tests/test_api.py
joshvillbrandt: Ah. Good call. Fixed!

PR #241, opened 2018-07-19
cache serialized objects within a single context

ryochiji:
Currently, we potentially serialize the same object over and over again when the same object instance appears many times (usually as sideloads).

For example, consider:
`/objectives/?include[]=subject.&include[]=rubric.rubric_items.`

There are actually only a handful of distinct subjects and a couple of sets of rubric/rubric items, but we re-serialize the subjects and rubrics for each objective. This PR introduces a cache that is tied to a specific serializer instance (which is shared within a single request context) to prevent such redundant serializations. (For that specific example above in our backend, this change reduces CPU usage by about half.)

### DIFF
--- a/dynamic_rest/conf.py
+++ b/dynamic_rest/conf.py
@@ -17,6 +17,12 @@ DYNAMIC_REST = {
     # ENABLE_SERIALIZER_CACHE: enable/disable caching of related serializers
     'ENABLE_SERIALIZER_CACHE': True,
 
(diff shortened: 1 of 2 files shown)


ryochiji:
@DeviateFish can you give me more specific feedback on the diff?

ryochiji:
Ah yeah, it's actually a much smaller change than it appears.

PR #243, opened 2018-07-23
Update DynamicListSerializer

ben-prezola:
Cast lookup key to string to allow for UUID primary keys

Refs #232


aleontiev:
:+1: thanks @ben-prezola

PR #249, opened 2018-08-28
Some upgrades to the CircleCI config

atzannes:
Mainly:
- Caching of the PyEnv step which takes 8 minutes
- Don't install Python 3.3 which is deprecated, instead install Python 3.6


jvineet:
Looks great. We should also add this caching update to altkube CircleCI config.

Issue #247, opened 2018-08-20
Unable to sort and filter JSON fields


dalenhughes:
I'm looking to expand the functionality to handle sorting and querying JSON field data. 
The PostgresSQL database uses a jsonb object which allows indexing. I would need to migrate one of the models in this project to include a doubly nested JSON Field. Is this possible with the current testing framework?

https://docs.djangoproject.com/en/2.1/ref/contrib/postgres/fields/

dalenhughes:
@ryochiji

joshvillbrandt:
Ping @ryochiji, @atzannes, @DeviateFish - would someone mind reviewing our request? We'd like to get this merged so that we aren't maintaining a separate fork.

Thanks for your time and support!

atzannes:
Hey @joshvillbrandt and @dalenhughes, thank you for your interest in the project. Unfortunately, I'm not very knowledgeable and will defer to Ryo. I contacted him and it's on his radar to reply to you as soon as he finds some time.

joshvillbrandt:
Awesome. Thanks for the update! :)

mattmendivil:
@aleontiev @suavesav and @chdastolfo I have a draft PR up for fixing this issue- I'm having some issues running tests locally so I'd like to utilize the workflows that are automatically run on PRs.

Could one of you approve running the workflow on this PR?

Thanks!

PR #250, opened 2018-08-31
allow for sorting/filtering of JSON objects when using PostgreSQL #247

dalenhughes:


DeviateFish:
Just a quick heads up on this... Our lead maintainer is currently on vacation, so it might be some time before this gets a proper review.

PR #257, opened 2018-10-15
Pass queries on JSON fields direct to Django

PaulWay:
Django allows queries to JSON fields to contain arbitrary sub-expressions that are looked up from the JSON content structure.  These queries look like related field lookups to dynamic-rest.  This patch finds queries on JSON fields and stops related field processing, instead passing the query to Django.


aleontiev:
Closing in favor of #260 as a more general solution

PR #260, opened 2018-10-24
add support for Transform on fields (year, month, day, etc)

ornoone:
last support for these Transform was to treat them as operator, but
it prevented the use of transform AND operator (__year__gt=2018)

now, if a filter on a field hit a non-nested field, instead of raising
a validation error, it treat the spare parts as Transform and add them
to the final query. it's more flexible and allow support for custom
Transform along with support for JSONField

### DIFF
--- a/dynamic_rest/filters.py
+++ b/dynamic_rest/filters.py
@@ -122,17 +122,15 @@ class FilterNode(object):
             # and field.source would be None
             rewritten.append(model_field_name)
 
-            if i == last:
-                break
-
             # Recurse into nested field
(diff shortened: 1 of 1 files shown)


ornoone:
this fix the same problem as #257 but this fix is far more generic.

PaulWay:
Looks good to me!

aleontiev:
:+1: from me and I don't see this breaking backwards compatibility
@ryochiji any concerns?

ornoone:
is there a reason for this PR to be blocked ?

PR #265, opened 2018-11-29
new viewset feature: patch-all

aleontiev:


dynamic_rest/viewsets.py
```diff
@@ -418,36 +441,150 @@ def _bulk_update(self, data, partial=False):
...
+        fields = serializer.get_all_fields()
+        validated = {}
+        for name, value in six.iteritems(data):
+            field = fields.get(name, None)
+            if field is None:
+                raise ValidationError(
+                    'Unknown field: "%s"' % name
+                )
+            source = field.source or name
+            if source == '*':
```
ryochiji: Do we also need to check for `read_only` or would that happen upstream somewhere?

tests/viewsets.py
```diff
@@ -151,6 +151,7 @@ class DogViewSet(DynamicModelViewSet):
     model = Dog
     serializer_class = DogSerializer
     queryset = Dog.objects.all()
+    ENABLE_PATCH_ALL = True
```
ryochiji: Should we disable by default?

ryochiji: nm

dynamic_rest/viewsets.py
```diff
@@ -418,36 +441,150 @@ def _bulk_update(self, data, partial=False):
...
+                'fur': 'gold'
+            }
+        """  # noqa
         if self.ENABLE_BULK_UPDATE:
-            partial = 'partial' in kwargs
-            bulk_payload = self._get_bulk_payload(request)
-            if bulk_payload:
-                return self._bulk_update(bulk_payload, partial)
+            patch_all = self.get_request_patch_all()
+            if (
```
ryochiji: style: line wrapping probably not necessary here

aleontiev: :+1:, I think this initially had longer names and one extra clause so it was too long to fit on one line

dynamic_rest/viewsets.py
aleontiev: Good catch, should check for it here

PR #271, opened 2018-12-31
Leave rest_framework router is unmodified

impala2:
Right now it's doing a shallow copy of the routes and hence your monkeypatch affects the source of the data. This causes all routes to accept `PATCH` and `DETELE` requests which in turn causes this exception instead of returning a 405:

> AssertionError: Expected view {viewset} to be called with a URL keyword argument named "pk". Fix your URL conf, or set the `.lookup_field` attribute on the view correctly.


aleontiev:
:+1: thanks @impala2 !

PR #274, opened 2019-01-23
remove dependency links, fix bulk update string processing, return 400s on update-without-id

aleontiev:


dynamic_rest/viewsets.py
```diff
@@ -582,8 +582,18 @@ def update(self, request, *args, **kwargs):
                     return self._bulk_update(bulk_payload, partial)
 
         # singular update
-        return super(DynamicModelViewSet, self).update(request, *args,
-                                                       **kwargs)
+        try:
+            return super(DynamicModelViewSet, self).update(request, *args,
+                                                           **kwargs)
+        except AssertionError as e:
+            if 'lookup_field' in str(e):
```
ryochiji: Where are these errors raised? Wondering if this is a safe assumption, and/or if the errors is from within DREST we could raise a custom error (but maybe it's from DRF/Django?)

aleontiev: @ryochiji I have updated the check to use a more reliable substring of the DRF assertion, changed the error message of the 400 to be more descriptive, and left a comment about the reasoning behind this change (DREST's support for bulk operations means that an update-without-ID should NOT be considered a misconfiguration as it is in DRF)

Makefile
```diff
@@ -41,11 +41,11 @@ install: $(INSTALL_DIR)
 # Install/update dependencies
 # Runs whenever the requirements.txt file changes
 $(INSTALL_DIR): $(INSTALL_DIR)/bin/activate
-$(INSTALL_DIR)/bin/activate: requirements.txt install_requires.txt dependency_links.txt
+$(INSTALL_DIR)/bin/activate: requirements.txt install_requires.txt
 	$(call header,"Updating dependencies")
 	@test -d $(INSTALL_DIR) || virtualenv $(INSTALL_DIR)
 	@$(INSTALL_DIR)/bin/pip install -q --upgrade pip setuptools flake8==2.4.0
-	@$(INSTALL_DIR)/bin/pip install --process-dependency-links -Ur requirements.txt
+	@$(INSTALL_DIR)/bin/pip install -U -r requirements.txt
```
aleontiev: Driveby -- we dont need dependency links, and they are gone in pip 19

PR #281, opened 2019-05-24
add Python 3.7 + Django 2.2

ryochiji:


aleontiev:
@ryochiji was this failing because of the `djay` 3.x issues?

ryochiji:
@aleontiev - Looking at the build failures, I don't think `djay` was the issue (or, at least, it was failing before it got to that step). Going to close anyhow since I got that other PR in.

PR #283, opened 2019-05-28
Add 'iexact' to available filters

vmihalachi:
Hi, I have added 'iexact' to the list of available filters. 

https://docs.djangoproject.com/en/2.1/ref/models/querysets/#iexact


aleontiev:
:+1:

PR #286, opened 2019-07-29
fix bug where implicit prefetch and explicit prefetch behaviors were …

ryochiji:
…inconsistent. Specifically, in the implicit prefetch scenario, the related serializer's `filter_queryset()` method wasn't being respected (because without that related field being included, part of the codepath didn't recognize it as a valid field).

### DIFF
--- a/tests/serializers.py
+++ b/tests/serializers.py
@@ -59,9 +59,6 @@ class LocationSerializer(DynamicModelSerializer):
             'cats', 'friendly_cats', 'bad_cats'
         )
 
-    def filter_queryset(self, query):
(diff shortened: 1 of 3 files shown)


tests/serializers.py
```diff
@@ -59,11 +59,15 @@ class Meta:
             'cats', 'friendly_cats', 'bad_cats'
         )
 
+    def filter_queryset(self, query):
```
DeviateFish: Don't we usually define functions after fields in serializers?  As a style thing, that is.

ryochiji: Yep, you're right. Moved.

PR #293, opened 2019-12-04
Update django to latest patch versions

simkimsia:


simkimsia:
How do I run the test to see if this will pass in circleci?

Or I have to do it on my own circleci account?

PR #297, opened 2020-01-24
Ensure DynamicRelationField.to_representation is passed correct instance

fdintino:
This fixes a regression that was originally introduced in d0c31d3 and a869d28

There were a few places in our code where we have subclassed `DynamicRelationField` and overridden `to_representation`, conditionally short-circuiting the serialization of a many to many field depending on the request, context, or a property of the parent instance. We had been relying on the fact that `to_representation` for a `ModelField` and a `DynamicRelationField` was always passed the parent instance. But in d0c31d3, and then a869d28, some of the responsibilities of `to_representation` were moved over to `get_attribute`.

After updating to the latest version our methods no longer worked, because `to_representation` was now being passed the `ManyRelatedManager`, not the parent instance. This seems to me like a bug, as it breaks expectations about the arguments and return values of `get_attribute` and `to_representation`. And so far as I can tell, moving the logic back from `get_attribute` doesn't break anything in dynamic-rest, since the two methods are only ever called together, e.g.: in `WithDynamicSerializerMixin._faster_to_representation()`:

```python
attribute = field.get_attribute(instance)
# ...
ret[field.field_name] = field.to_representation(attribute)
```

I've included a test that shows roughly how we sometimes override `to_representation`.

### DIFF
--- a/dynamic_rest/fields/fields.py
+++ b/dynamic_rest/fields/fields.py
@@ -268,20 +268,7 @@ class DynamicRelationField(WithRelationalFieldMixin, DynamicField):
         )
 
     def get_attribute(self, instance):
-        serializer = self.serializer
-        model = serializer.get_model()
-
-        # attempt to optimize by reading the related ID directly
-        # from the current instance rather than from the related object
-        if not self.kwargs['many'] and serializer.id_only():
-            return instance
-        elif model is not None:
-            try:
-                return getattr(instance, self.source)
-            except model.DoesNotExist:
-                return None
-        else:
-            return instance
+        return instance
 
     def to_representation(self, instance):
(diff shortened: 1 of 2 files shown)


fdintino:
This PR was closed accidentally. Re-opening.

fdintino:
Any chance this might get a review and merged? We're currently installing from our github fork, but we'd prefer to avoid that if we can. Thanks!

defbyte:
Seconding the request for eyes on this, it would be super helpful ❤️

fdintino:
Let me know if there's anything I can to do to help get this merged in. We'd prefer it if we didn't need to maintain a fork. Thank you!

simkimsia:
@suavesav can help?

PR #300, opened 2020-04-13
Update inflection to 0.4.0

simkimsia:


simkimsia:
Ok to merge?

install_requires.txt
```diff
@@ -1,5 +1,5 @@
 Django>=1.8,<2.0
 djangorestframework>=3.1.0,<3.10
-inflection==0.3.1
```
suavesav, line 3: Looks like inflection 0.4.0 requires python>=3.5. since this `install_requires` file gets used by python2, we should not be upgrading the version of inflection here (this file is getting deprecated in the next release anyway)

simkimsia: So I should hold off until the install requires is completely removed or ?

suavesav: Its been removed. Feel free to rebase

PR #302, opened 2020-05-13
Update tox testing

suavesav:
Various updates to automated testing through tox:
 * Reduce tox testing output in circle
 * Update python minor versions, add 3.8
 * Update tox version
 * Add classifiers to `setup.py`
 * Update django, drf minor versions to get tested against in tox


.circleci/python-versions.txt
```diff
@@ -1 +1 @@
-2.7.16 3.5.7 3.6.9 3.7.4
+2.7.17 3.5.8 3.6.9 3.7.5 3.8.0
```
ryochiji, line 1: You added 3.8 here but not in tox.ini. Is that intentional?

suavesav: Kind of. When I added this to `tox` some tests were failing, so we are not 3.8 ready yet.

However keeping this in here speeds up circle testing in the future since we don't have to rebuild the python version cache when we add it back in. Just saving 5 future minutes by leaving this in 😬

PR #303, opened 2020-05-13
Drop support for py2 and move to Github Actions

suavesav:
Drop support for py2 in newer version of dynamic-rest.
 * Also simplify testing and get rid of requirements.test.txt

Additionally this PR also drops CircleCI in favor of Github Actions. There are a few reasons for this
1. Github Actions is free for public repositories: https://github.com/pricing
2. Testing in CircleCI is done within tox, aka a venv inside a venv which abstracts the actual testing environment down too far. In Github Actions we create a different container for each permutation we want to test. I tried to do a similar setup in Circle, but they don't really support it afaict: https://circleci.com/blog/continuous-deployment-of-gatsby-apps-to-heroku/
3. Github Actions allows for parallelism of up to 20 which means tests can complete in <4 mins instead of the 10-15 mins we see on circle: https://help.github.com/en/actions/getting-started-with-github-actions/about-github-actions

Unfortunately, translating the integration tests from circle to github actions was [causing an issue](https://github.com/AltSchool/dynamic-rest/actions/runs/109854145) I could not resolve within a timeboxed period of 1 day. So we are leaving the integration test in Circle.

Regardless, I think the time-savings and cost-savings on CICD are worth it.


suavesav:
Waiting on a fix to `djay`

requirements.txt
```diff
@@ -1,12 +1,12 @@
-Sphinx==1.6.5
-dj-database-url==0.4.2
-flake8==2.4.0
+dj-database-url==0.5.0
```
aleontiev, line 1: Might be worth switching this project to `poetry` soon to reduce these files:
- `requirements.txt`
- `setup.py`
- `requirements.test.txt`
- `install_requires.txt`
Down to just:
- `pyproject.toml` 

The project should still be buildable by `pip`/`pipenv`/etc as a dependency without poetry

aleontiev:
@suavesav regarding the Github Actions issue, I couldn't find any logs from your link to the issue; is that part of the problem, or are the logs gone now?

PR #304, opened 2020-06-23
[ALTOS-22468] add fix for sort by relation

aleontiev:
When sorting by a related many-field, we need to call `distinct()` in order to remove duplicates


aleontiev:
@ryochiji FYI

PR #308, opened 2020-07-13
bump major version number since dropping py2 is a breaking change

suavesav:
https://github.com/AltSchool/dynamic-rest/pull/303 would cause this lib to break for people using py2, so bump the major version number prior to releasing


aleontiev:
:+1:

PR #318, opened 2021-01-06
Update install_requires.txt

simkimsia:
to allow any patch version of 3.1


suavesav:
Fixed in https://github.com/AltSchool/dynamic-rest/pull/331

PR #331, opened 2021-11-24
Update installation requirements to actually allow installation along…

chdastolfo:
…side Django 3.2


suavesav (approved):
Approved pending CI passes

PR #336, opened 2022-06-14
add new "exclude count" page number pagination option

aleontiev:
...to support a faster page-number pagination over large datasets

- Like standard `PageNumberPagination`, except the results are never counted
- Uses the N+1 trick from `CursorPagination` to determine whether the dataset has more pages
- Can be enabled on any request via query parameter `exclude_count` (name can be configured)

TODO: 
- [x] ensure all current tests pass
- [x] add new test case with `exclude_count` enabled

### DIFF
--- a/dynamic_rest/paginator.py
+++ b/dynamic_rest/paginator.py
@@ -1,3 +1,6 @@
+# adapter from Django's django.core.paginator
+# adds support for the "exclude_count" parameter
+
 from math import ceil
 
 from django.utils.functional import cached_property
@@ -58,7 +61,7 @@ class DynamicPaginator(Paginator):
         bottom = (number - 1) * self.per_page
         top = bottom + self.per_page
         if self.exclude_count:
-            # fetch one extra item to determine if more pages are available
+            # always fetch one extra item to determine if more pages are available
             top = top + 1
         else:
             # skip validating against count
--- a/dynamic_rest/paginator.py
+++ b/dynamic_rest/paginator.py
@@ -1,0 +1,94 @@
+from math import ceil
+
+from django.utils.functional import cached_property
+from django.core.paginator import Paginator
+
+try:
+    from django.utils.translation import gettext_lazy as _
+except ImportError:
+    def _(x):
+        return x
+
+
+
+class InvalidPage(Exception):
+    pass
+
+
+class PageNotAnInteger(InvalidPage):
+    pass
+
+
+class EmptyPage(InvalidPage):
+    pass
+
+
+class DynamicPaginator(Paginator):
+
+    def __init__(self, object_list, per_page, orphans=0,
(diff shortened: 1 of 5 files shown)


chdastolfo (approved):
LGTM post version bump

dynamic_rest/paginator.py
```diff
@@ -0,0 +1,96 @@
...
+
+class DynamicPaginator(Paginator):
+
+    def __init__(self, object_list, per_page, orphans=0,
+                 allow_empty_first_page=True, exclude_count=False):
+        self.object_list = object_list
+        self._check_object_list_is_ordered()
+        self.per_page = int(per_page)
+        self.orphans = int(orphans)
+        self.allow_empty_first_page = allow_empty_first_page
```
suavesav, line 38: can we just call the super class' __init__ to do this?

dynamic_rest/paginator.py
```diff
@@ -0,0 +1,96 @@
...
+        self.exclude_count = exclude_count
+
+    def validate_number(self, number):
```
suavesav, line 41: what version of django did you use to get these method's django defn? maybe we should add that info here as a comment?

dynamic_rest/paginator.py
```diff
@@ -0,0 +1,96 @@
...
+        """Return a Page object for the given 1-based page number."""
+        number = self.validate_number(number)
+        bottom = (number - 1) * self.per_page
+        top = bottom + self.per_page
+        if self.exclude_count:
+            # always fetch one extra item
+            # to determine if more pages are available
+            top = top + 1
+        else:
+            # skip validating against count
```
suavesav, line 69: this comment line should be in the `if` part of this conditional or can be removed altogether

suavesav (commented):
Couple of questions about DynamicPaginator

dynamic_rest/paginator.py
aleontiev: Good point

dynamic_rest/paginator.py
aleontiev: This was taken from 3.1, but it seems to work across all tested versions in DREST tox (from 2.2 onwards)

dynamic_rest/paginator.py
aleontiev: done

dynamic_rest/paginator.py
aleontiev: done

dynamic_rest/paginator.py
aleontiev: done

aleontiev:
@chdastolfo @suavesav 
- all tests pass including the new one I added 
- addressed all PR feedback
- bumped version to `2.1.3` in setup.py

so I am going to merge 🚀

PR #349, opened 2023-03-24
Feature/advanced filter

aleontiev:


dynamic_rest/filters.py
```diff
@@ -130,16 +137,46 @@ def generate_query_key(self, serializer):
...
+    q = {}
+    if key.startswith('-'):
+        negate = True
+        key = key[1:]
+    parts = key.split('.')
+    operator = 'eq'
+    if parts[-1] in DynamicFilterBackend.VALID_FILTER_OPERATORS:
+        operator = parts.pop()
+    if operator == 'eq':
+        operator = None
```
sjoanes, line 171: why not just initialize this to `None`

dynamic_rest/filters.py
```diff
@@ -298,39 +340,50 @@ def _filters_to_query(self, includes, excludes, serializer, q=None):
...
+            '_complex' not in filters
+        ):
+            includes = filters.get('_include')
+            excludes = filters.get('_exclude')
+            q = q or Q()
 
-        if not includes and not excludes:
-            return None
+            if not includes and not excludes:
+                return None
```
sjoanes, line 351: curious if None is ok here or if it should be `Q()`

sjoanes: nvm this is from before

dynamic_rest/filters.py
```diff
@@ -513,67 +545,61 @@ def _build_queryset(
...
+                    for field in implicitly_included
+                    if field in all_fields
+                }
+            )
 
+        nested_filters = None
         if filters is None:
             filters = self._get_requested_filters()
+            nested_filters = (
+                self._get_requested_filters(nested=True)
```
sjoanes, line 568: #nitpick can we simplify this somehow? calling `_get_requested_filters` to call it again with `nested=True` feels awk

dynamic_rest/filters.py
```diff
@@ -393,10 +433,7 @@ def _build_requested_prefetches(
 
             source = field.source or name
             if '.' in source:
-                raise ValidationError(
-                    'nested relationship values '
-                    'are not supported'
-                )
+                raise ValidationError('nested relationship values ' 'are not supported')
```
sjoanes, line 436: weird string

dynamic_rest/filters.py
aleontiev: In case it was explicitly set to `eq` by the user, its still better to leave it off because some types don't work correctly with `eq` (like UUID)

dynamic_rest/filters.py
aleontiev: Yeah.. it would be ideal if all of the filters were consolidated in one tree as they are before this change. I can see about consolidating them back into one object

aleontiev: @sjoanes done, looks less awk now

sjoanes (approved):
LGTM

PR #357, opened 2023-09-16
Update compatibility

twine003:
- Not longuer suported NullBooleanField in newers versions of Django.
- Updated compatibility to djangorestframework 3.17

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -1,9 +1,9 @@
 from setuptools import find_packages, setup
 
-NAME = 'dynamic-rest'
-DESCRIPTION = 'Dynamic API support to Django REST Framework.'
+NAME = 'dynamic-rest-4'
+DESCRIPTION = 'Dynamic API support to Django REST Framework, Compatible with Django 4'
 URL = 'http://github.com/twine003/dynamic-rest'
-VERSION = '3.1.0'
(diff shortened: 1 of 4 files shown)


BillSchumacher:
If anyone is looking for an installable package with some other improvements I have `dynamic-rest-bse` on pypi.

https://github.com/BillSchumacher/dynamic-rest if you would like to actively improve this package.

Support for Python 2 and Django < 3 is dropped.

setup.py
```diff
@@ -1,9 +1,9 @@
 from setuptools import find_packages, setup
 
-NAME = 'dynamic-rest'
-DESCRIPTION = 'Dynamic API support to Django REST Framework.'
-URL = 'http://github.com/AltSchool/dynamic-rest'
-VERSION = '2.1.7'
+NAME = 'dynamic-rest-4'
+DESCRIPTION = 'Dynamic API support to Django REST Framework, Compatible with Django 4'
+URL = 'http://github.com/twine003/dynamic-rest'
```
DuncanTormey, line 5: I think this might need to change back to AltSchool

suavesav:
We have django 4.2 compatibility in v2.2.1. (updated in pypi as well)

PR #374, opened 2024-08-23
Drop "six" package

gorodovoy9000:
Since dynamic-rest is using modern Py3 interpreters - we can now remove "six" package from lib and replace it with Py3 methods

plus: fixed some potent bug in file dynamic_rest/routers.py line 332
fields = getattr(serializer, 'get_link_fields', lambda: [])() to
fields = getattr(serializer, 'get_link_fields', lambda: {})()
because further line 340 use "fields" as dict

### DIFF
--- a/dynamic_rest/datastructures.py
+++ b/dynamic_rest/datastructures.py
@@ -1,5 +1,4 @@
 """This module contains custom data-structures."""
-import six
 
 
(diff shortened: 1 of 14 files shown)


auvipy (commented):
lgtm

